PR #4, opened 2019-01-07
tickets/DM-17060

natelust:
Add Switch for setting Datatype name templates

### DIFF
--- a/python/lsst/ctrl/mpexec/cmdLineParser.py
+++ b/python/lsst/ctrl/mpexec/cmdLineParser.py
@@ -99,6 +99,7 @@ _ACTION_MOVE_TASK = _PipelineActionType("move_task", r"(?P<label>.+):(?P<value>-
 _ACTION_LABEL_TASK = _PipelineActionType("relabel", "(?P<label>.+):(?P<value>.+)")
 _ACTION_CONFIG = _PipelineActionType("config", "(?P<label>.+):(?P<value>.+=.+)")
 _ACTION_CONFIG_FILE = _PipelineActionType("configfile", "(?P<label>.+):(?P<value>.+)")
+_ACTION_NAME_TEMPLATES = _PipelineActionType("name_templates", "(?P<label>[^.]+):(?P<value>.+)")
 
 
 class _LogLevelAction(Action):
@@ -389,6 +390,10 @@ def makeParser(fromfile_prefix_chars='@', parser_class=ArgumentParser, **kwargs)
         subparser.add_argument("-C", "--configfile", metavar="LABEL:PATH",
                                dest="pipeline_actions", action='append', type=_ACTION_CONFIG_FILE,
                                help="Configuration override file(s), applies to a task with a given label.")
+        subparser.add_argument("--dataset-name-substitution", metavar="LABEL:VALUE",
+                               dest='pipeline_actions', action='append', type=_ACTION_NAME_TEMPLATES,
+                               help='Configuration name template values, applies to a task with a given '
+                                    'label.')
         subparser.add_argument("-o", "--order-pipeline", dest="order_pipeline",
                                default=False, action="store_true",
                                help="Order tasks in pipeline based on their data dependencies, "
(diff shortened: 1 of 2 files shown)


python/lsst/ctrl/mpexec/cmdLineParser.py
```diff
@@ -381,6 +382,10 @@ def makeParser(fromfile_prefix_chars='@', parser_class=ArgumentParser, **kwargs)
         subparser.add_argument("-C", "--configfile", metavar="LABEL:PATH",
                                dest="pipeline_actions", action='append', type=_ACTION_CONFIG_FILE,
                                help="Configuration override file(s), applies to a task with a given label.")
+        subparser.add_argument("--dataset-name-substitution", metavar="LABEL:VALUE",
+                               dest='pipeline_actions', action='append', type=_ACTION_NAME_TEMPLATES,
+                               help='Configuration name template values, applies to a task with a given '
+                                    'label.')
```
andy-slac: Looking at the code I think VALUE in this case is a Python mapping of string to string. It may be rather messy to write that on command line and will need a lot of quoting, I suppose it should look like:

    --dataset-name-substitution 'FancyTask:{"inputCoaddName": "deep2", "outputCoaddName": "deep2"}'

Could some other format with less quoting be more user-friendly, maybe something like:

    --dataset-name-substitution FancyTask:inputCoaddName=deep2,outputCoaddName=deep2

python/lsst/ctrl/mpexec/cmdLineParser.py
```diff
@@ -91,6 +91,7 @@ def __call__(self, value):
 _ACTION_LABEL_TASK = _PipelineActionType("relabel", "(?P<label>.+):(?P<value>.+)")
 _ACTION_CONFIG = _PipelineActionType("config", "(?P<label>[^.]+)[.](?P<value>.+=.+)")
 _ACTION_CONFIG_FILE = _PipelineActionType("configfile", "(?P<label>.+):(?P<value>.+)")
+_ACTION_NAME_TEMPLATES = _PipelineActionType("name_templates", "(?P<label>[^.]+)[=](?P<value>.+)")
```
andy-slac: Should this use `:` instead of `=` for consistency with other options?

python/lsst/ctrl/mpexec/cmdLineParser.py
natelust: Jim and I debated this a little bit with pros and cons for several different ways. The conclusion we reached is that for now we will leave it as is, and collect data from several people in the group for the input pattern they would prefer, and make any changes in a future ticket.

python/lsst/ctrl/mpexec/cmdLineParser.py
natelust: I really have no opinion on this, I had thought having the equals sign make it more akin to setting a config option, as you are kind of assigning something to that task, but then again it is a different pattern because normally the label is just that a label not something that is assignable. So maybe it would be good to go back to : as the command line switch name already indicates that you are doing an action.

python/lsst/ctrl/mpexec/cmdLineParser.py
```diff
@@ -91,6 +91,7 @@ def __call__(self, value):
 _ACTION_LABEL_TASK = _PipelineActionType("relabel", "(?P<label>.+):(?P<value>.+)")
 _ACTION_CONFIG = _PipelineActionType("config", "(?P<label>[^.]+)[.](?P<value>.+=.+)")
 _ACTION_CONFIG_FILE = _PipelineActionType("configfile", "(?P<label>.+):(?P<value>.+)")
+_ACTION_NAME_TEMPLATES = _PipelineActionType("name_templates", "(?P<label>[^.]+):(?P<value>.+)")
```
andy-slac: That may not work too well if `value` contains `:` (as it should for dictionary). regexp is greedy by default and will match `:` with rightmost `:` in the string. Having unit test would probably catch it.

PR #5, opened 2019-01-09
Parse dictionary in command line parser (DM-17067)

andy-slac:
Now the value of --dataset-name-substitution is parsed directly during
command line parsing. Added unit tests for that new feature too.
Fix parsing of dataset-name-substitution option.

### DIFF
--- a/python/lsst/ctrl/mpexec/cmdLineParser.py
+++ b/python/lsst/ctrl/mpexec/cmdLineParser.py
@@ -99,7 +99,7 @@ _ACTION_MOVE_TASK = _PipelineActionType("move_task", r"(?P<label>.+):(?P<value>-
 _ACTION_LABEL_TASK = _PipelineActionType("relabel", "(?P<label>.+):(?P<value>.+)")
 _ACTION_CONFIG = _PipelineActionType("config", "(?P<label>.+):(?P<value>.+=.+)")
 _ACTION_CONFIG_FILE = _PipelineActionType("configfile", "(?P<label>.+):(?P<value>.+)")
-_ACTION_NAME_TEMPLATES = _PipelineActionType("name_templates", "(?P<label>[^.]+):(?P<value>.+)")
+_ACTION_NAME_TEMPLATES = _PipelineActionType("name_templates", "(?P<label>[^:]+):(?P<value>.+)")
 
 
 class _LogLevelAction(Action):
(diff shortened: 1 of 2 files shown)


python/lsst/ctrl/mpexec/cmdLineParser.py
```diff
@@ -93,13 +94,32 @@ def __repr__(self):
         return f"_PipelineActionType(action={self.action})"
 
 
+def _nameTemplatesType(value):
+    """Convert name_templates option value to a dictionary.
+    """
+    # try to parse value as python literal expression
+    parsedNamesDict = ast.literal_eval(value)
```
natelust: You may want to catch any errors thrown here by literal_eval, during parsing, as a mistake causes an error message to the user that is very confusing, instead of helping them figure out where there actual problem is at.

andy-slac: Errors (exceptions) that happen here are handled by `argparse` and it does not really matter what is thrown here, `argparse` will ignore exception type/value and will print the same message, e.g.:
```
 pipetask -b ../ci_hsc/DATA/butler.yaml -i shared/ci_hsc -o out1 run -t multiBand.DetectCoaddSourcesTask -t mergeDetections.MergeDetectionsTask:mdt --dataset-name-substitution 'mdt:XXX'
usage: pipetask run [-h] [-g PATH] [--skip-init-writes] [--init-only]
                    [--register-dataset-types] [-p PATH] [-t TASK[:LABEL]]
                    [-d LABEL] [-m LABEL:NUMBER] [-l LABEL:NEW_LABEL]
                    [-c LABEL:NAME=VALUE] [-C LABEL:PATH]
                    [--dataset-name-substitution LABEL:VALUE] [-o]
                    [--skip-existing] [-s PATH] [-q PATH]
                    [--pipeline-dot PATH] [--qgraph-dot PATH]
                    [--show ITEM|ITEM=VALUE]
pipetask run: error: argument --dataset-name-substitution: invalid _PipelineActionType(action=name_templates) value: 'mdt:XXX'
```
We can probably replace that `_PipelineActionType(action=name_templates)` with something more user-friendly if people think it's too confusing.

natelust: oh no, I think that is fine, it's what I had in mind. I didn't realize argparse was catching it. Everything looks good to me then

PR #15, opened 2019-03-19
tickets/DM-18367 Get butler working with Oracle

natelust:
Close the registry before exiting the activator

Close the registry before exiting the activator to avoid an
exception being raised in the destructor of the sqlalchemy
connection

### DIFF
--- a/python/lsst/ctrl/mpexec/cmdLineFwk.py
+++ b/python/lsst/ctrl/mpexec/cmdLineFwk.py
@@ -27,6 +27,7 @@ __all__ = ['CmdLineFwk']
 # -------------------------------
 #  Imports of standard modules --
 # -------------------------------
+import contextlib
 import fnmatch
 import logging
 import pickle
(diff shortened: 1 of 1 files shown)


python/lsst/ctrl/mpexec/cmdLineFwk.py
```diff
@@ -399,6 +399,8 @@ def runPipeline(self, graph, taskFactory, args):
         if not args.init_only:
             executor = MPGraphExecutor(numProc=args.processes, timeout=self.MP_TIMEOUT)
             executor.execute(graph, butler, taskFactory)
+        # Explicitly shutdown the registry
+        butler.registry.shutdown()
```
andy-slac: I don't think I like to see this here. My idea is that this code did nothing to create/setup the registry so it also should not be responsible for shutting it down. You already defined `Butler.__del__` so I think that should take care of the registry life time (and see my other comments to, I think we should not be exposing this method at all and let registry handle connection closing).

natelust: See my comment on the other ticket, but I tried that first and it does not work. More so, even letting the butlers __del__ fire does not solve the issue of an exception being raised in the destruction of the registry. The only way I found to avoid the exception was to explicitly close the registry. This code did create the butler which creates the registry, if you would prefer I add a shutdown method to the butler which chain calls this I will so that the butler create and close is symmetric. In this case I think letting __del__ fire might not be enough due to some references that may be kept by the multiprocessing module. Also keep in mind per the python documentation there is no gaurentee on when __del__ will be fired and the order in which objects are garbage collected is not deterministic. I think this is what is causing issues in the finalization of the sqlachemy objects.

timj: I think a butler close method is better than external users calling registry close.

timj: couldn't this also be a butler context handler?

natelust: The connection object and registry objects are held for the entire lifetime of the butler object, and so I dont see how this would work with a context manager.

timj: Ok. I only looked at the code quickly and saw the butler creation a few lines before the shutdown.

python/lsst/ctrl/mpexec/cmdLineFwk.py
```diff
@@ -399,6 +399,10 @@ def runPipeline(self, graph, taskFactory, args):
         if not args.init_only:
             executor = MPGraphExecutor(numProc=args.processes, timeout=self.MP_TIMEOUT)
             executor.execute(graph, butler, taskFactory)
+        # Explicitly shutdown the butler as automatic collection can
+        # sometimes lead to an exception being thrown on exit. This
+        # does not affect processing but is not friendly to the user.
+        butler.shutdown()
```
andy-slac: Moving this call to butler instead of registry is certainly better if we can't get rid of it completely. Would it better to call it `close()` instead of `shutdown()`? We could then use `contextlib.closing` for context handling as Tim proposed.

PR #16, opened 2019-03-25
DM-17932: Validate object type when reading objects from pickle

andy-slac:
Trivial change to the code. Also adds less trivial unit test for these
checks to validate validation. CmdLineFwk unit tests will be expanded
later with more tests for other methods (some will need reasonable
registry contents).

### DIFF
--- a/tests/test_cmdLineFwk.py
+++ b/tests/test_cmdLineFwk.py
@@ -1,0 +1,216 @@
+# This file is part of ctrl_mpexec.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+"""Simple unit test for cmdLineFwk module.
+"""
+
+import argparse
+import contextlib
+import pickle
+import os
+import tempfile
+import unittest
+
+from lsst.ctrl.mpexec.cmdLineFwk import CmdLineFwk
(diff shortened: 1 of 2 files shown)


tests/test_cmdLineFwk.py
```diff
@@ -0,0 +1,216 @@
...
+from lsst.ctrl.mpexec.cmdLineFwk import CmdLineFwk
+from lsst.ctrl.mpexec.cmdLineParser import _PipelineAction
+import lsst.pex.config as pexConfig
+from lsst.pipe.base import (Pipeline, PipelineTask, PipelineTaskConfig,
+                            QuantumGraph, TaskFactory, InitInputDatasetField)
+import lsst.utils.tests
+
+
+@contextlib.contextmanager
+def makeTmpFile():
```
natelust: Is there a reason you chose to make your own Context manager instead of using tempfile.TemporaryFile ?

tests/test_cmdLineFwk.py
```diff
@@ -0,0 +1,216 @@
...
+            pipeline = fwk.makePipeline(taskFactory, args)
+            self.assertIsInstance(pipeline, Pipeline)
+            self.assertEqual(len(pipeline), 0)
+
+            # pickle with wrong object type
+            with open(tmpname, "wb") as pickleFile:
+                pickle.dump({}, pickleFile)
+            args = _makeArgs(pipeline=tmpname)
+            with self.assertRaises(TypeError):
+                pipeline = fwk.makePipeline(taskFactory, args)
```
natelust: Small nitpick, but a statement assigning this to a variable seems needless for this and the test below.

tests/test_cmdLineFwk.py
```diff
@@ -0,0 +1,216 @@
...
+        building.
+        """
+        fwk = CmdLineFwk()
+        taskFactory = TaskFactoryMock()
+
+        with makeTmpFile() as tmpname:
+
+            # make empty graph and store it in a file
+            qgraph = QuantumGraph()
+            with open(tmpname, "wb") as pickleFile:
```
natelust: Would this be better to do like the test above where you using the save the graph by using the save_graph argument to _makeArgs?

tests/test_cmdLineFwk.py
andy-slac: The reason is that I need the name of the file and not file object. It could potentially be done with `NamedTemporaryFile` but it still needs additional code to close the file, so I prefer old-fashioned  `mkstemp` instead in a separate context manager.

tests/test_cmdLineFwk.py
andy-slac: The `CmdLineFwk.makeGraph()` method cannot make empty QGraph, and to make non-empty one it needs butler which we do not have yet implemented for unit tests. So the only way to save QGraph to a pickle file now is to make an instance manually and write it ourselves. That is simple enough for this unit test.

PR #18, opened 2019-04-18
Issue a warning when generated QuantumGraph is empty (DM-19372)

andy-slac:


python/lsst/ctrl/mpexec/cmdLineFwk.py
```diff
@@ -364,6 +365,14 @@ def makeGraph(self, pipeline, taskFactory, args):
             graphBuilder = GraphBuilder(taskFactory, butler.registry, args.skip_existing)
             qgraph = graphBuilder.makeGraph(pipeline, coll, args.data_query)
 
+        # count quanta in graph and give a warning if it's empty
+        nQuanta = sum(1 for q in qgraph.quanta())
+        if nQuanta == 0:
+            warnings.warn("QuantumGraph is empty", stacklevel=2)
```
MichelleGower: Works great.  Does the output of warnings.warn also go through the logging module?

andy-slac: It can be (as you have found already) redirected to `logging` with `logging.captureWarnings`. We probably want to add this to `pipetask` as default settings, I think it's trivial change and can be done on this ticket too.

PR #20, opened 2019-06-11
DM-20054: Adapt to change in DatasetTypeDescriptor.

TallJimbo:
The datasetType property has been replaced by the makeDatasetType method, in order to ensure dimensions are normalized.


python/lsst/ctrl/mpexec/dotTools.py
```diff
@@ -223,6 +224,7 @@ def pipeline2dot(pipeline, file, taskFactory=None):
         close = True
 
     print("digraph Pipeline {", file=file)
+    universe = DimensionUniverse.fromConfig()
```
andy-slac: Could you move this line to the very beginning of the method? If it raises for some reason I'd prefer not to write anything to a file or not to open a file.

PR #23, opened 2019-08-15
DM-20205 Update to use new PipelineTask API

natelust:


python/lsst/ctrl/mpexec/dotTools.py
```diff
@@ -233,22 +233,17 @@ def pipeline2dot(pipeline, file, taskFactory=None):
...
 
-        # we will need task class for next operations, make sure it is loaded
-        taskClass = _loadTaskClass(taskDef, taskFactory)
-
-        # task inputs
-        dsMap = taskClass.getInputDatasetTypes(taskDef.config)
-        for dsTypeDescr in dsMap.values():
-            dsType = dsTypeDescr.makeDatasetType(universe)
+        for name in taskDef.connections.inputs:
+            attr = getattr(taskDef.connections, name)
```
andy-slac: I see a bunch of repetitions of this pattern:

        for name in taskDef.connections.<some_connections>:
            attr = getattr(taskDef.connections, name)

Would it be better to add a helper method for iteration?

natelust: added in pipe base

PR #30, opened 2019-10-31
DM-21889: Avoid option reuse in pipetask

andy-slac:
Single-letter options are made unique:
- `list` subcommand only uses long options now
- timeout option also uses long option
- `-p pipeline.pickle` is replaced with `-r pipeline.pickle`


andy-slac:
If there is another disruptive interface change then I think it makes sense to do it in one announcement rather than few close ones. @natelust, how close are you to finishing your fixes? Do you want me to delay merging this request into master so that both our merges will be close together? If you need to rebase you can use my branch and then rebase to master again one my branch merges.

natelust:
My changes should soon be ready to review. The scope of this work keeps creeping a bit. After today I hope to have everything in the state it will be in. I think it might be a good idea to sit down and talk about the changes that are coming. It would probably be best if you @TallJimbo and I were all there.

PR #33, opened 2019-11-11
DM-22166: Fix pipetask --show=pipeline option

andy-slac:
The code still expected pipeline to be a list of TaskDefs which is not
true anymore, need to call special method to get that list. Added unit
tests to check --show options and fixed all cases needed for correct
functioning.

### DIFF
--- a/python/lsst/ctrl/mpexec/cmdLineFwk.py
+++ b/python/lsst/ctrl/mpexec/cmdLineFwk.py
@@ -397,7 +397,7 @@ class CmdLineFwk:
                     continue
 
             if showCommand == "pipeline":
-                for taskDef in pipeline:
+                for taskDef in pipeline.toExpandedPipeline():
                     print(taskDef)
             elif showCommand == "config":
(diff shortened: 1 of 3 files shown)


python/lsst/ctrl/mpexec/cmdLineFwk.py
```diff
@@ -397,7 +397,7 @@ def showInfo(self, args, pipeline, graph=None):
                     continue
 
             if showCommand == "pipeline":
-                for taskDef in pipeline:
+                for taskDef in pipeline.toExpandedPipeline():
```
natelust: I think this would be better to just be print(pipeline), as it is a good way to see what the pipeline is expanded to if there are inherits, configs, contracts, etc. What this is doing now is really just showing the tasks that are in the pipeline.

andy-slac: Agree, that should look more informative, though it breaks compatibility (which I think we will survive).

PR #35, opened 2019-11-26
DM-22364: Now must explicitly specify where a fake posix datastore exists

timj:


tests/test_cmdLineFwk.py
```diff
@@ -126,10 +126,11 @@ def _makeArgs(pipeline=None, qgraph=None, pipeline_actions=(), order_pipeline=Fa
     pipeline_dot : `str`
         Name of the DOT file to write pipeline graph.
     qgraph_dot : `str`
-        Name of the DOT file to write QGrpah representation.
+        Name of the DOT file to write QGraph representation.
     """
     args = argparse.Namespace()
-    args.butler_config = None
+    args.butler_config = Config()
+    args.butler_config.configFile = "."  # Lie about where the datastore is located
```
andy-slac: I'm against lying in general. Could you rephrase it to something like "configFile must be specified but it does not matter in this case"? Could you specify `configFile` in a constructor? If empty `Config` constructor is not useful for cases like this maybe it should be disabled entirely?

timj: In this particular case you only want access to registry and don't care where datastore is located. An empty Config to construct a Butler in theory could be fine since defaults are picked up, but in the specific case the default is to use a relocatable datastore configuration so you then have a bit of a chicken and egg problem. I don't think it's the fault of Config that it's not usable to construct a Butler if it's empty.

andy-slac: OK to merge.

PR #36, opened 2019-12-06
DM-22162: Add metadata writing to PipelineTask execution logic (ctrl_mpexec) 

andy-slac:
Add support for writing task metadata to a special output dataset.


python/lsst/ctrl/mpexec/singleQuantumExecutor.py
```diff
@@ -231,3 +233,11 @@ def runQuantum(self, task, quantum):
         # Call task runQuantum() method. Any exception thrown by the task
         # propagates to caller.
         task.runQuantum(butlerQC, inputRefs, outputRefs)
+
+        if taskDef.metadataDatasetName is not None:
+            # Metadata is supposed to be of the PropertyList type, its dimensions
+            # correspond to a task quantum
+            dimensions = self.butler.registry.dimensions.extract(taskDef.connections.dimensions)
+            dsType = DatasetType(taskDef.metadataDatasetName, dimensions, "PropertyList")
```
natelust: Isn't this already calculated and added to the outputs in the quantum object?

andy-slac: I think you are right, I was struggling with how to find it in a Quantum, but I guess I can look it up by its name in `Quantum.outputs`. Will try that quickly.

andy-slac: I have re-implemented this with simple lookup in quantum outputs, thanks for catching this!

PR #39, opened 2019-12-12
DM-22504: Support for lsstDebug functionality in Gen3 middleware

andy-slac:
The --debug option existed but did not do anyhting, added code that
imports `debug.py` when that option is set. Compared to CmdLineTask in
`pipetask` I do it in two places, once before PreExecInit (in case
someone needs to debug task construction) and once before running each
task. Latter is not needed if `multiprocess` uses 'fork' start method
but it is impossible to guarantee that is always true on all platforms.

### DIFF
--- a/python/lsst/ctrl/mpexec/cmdLineFwk.py
+++ b/python/lsst/ctrl/mpexec/cmdLineFwk.py
@@ -358,6 +358,17 @@ class CmdLineFwk:
         if not butler.run:
             raise ValueError("no output collection defined in data butler")
 
+        # Enable lsstDebug debugging. Note that this is done once in the
+        # main process before PreExecInit and it is also repeated before
+        # running each task in SingleQuantumExecutor (which may not be
+        # needed if `multipocessing` always uses fork start method).
+        if args.enableLsstDebug:
+            try:
+                _LOG.debug("Will try to import debug.py")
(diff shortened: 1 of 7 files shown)


python/lsst/ctrl/mpexec/cmdLineFwk.py
```diff
@@ -358,6 +358,18 @@ def runPipeline(self, graph, taskFactory, args, butler=None):
...
 
+        # Enable lsstDebug debugging. Note that this is done once in the
+        # main process before PreExecInit and it is also repeated before
+        # running each task in SingleQuantumExecutor (which may not be
+        # needed if `multipocessing` always uses fork start method).
+        if args.enableLsstDebug:
+            try:
+                _LOG.debug("Will try to import debug.py")
+                import debug
+                assert debug  # silence pyflakes
```
TallJimbo: I think you mean "silence flake8", as that's the linter we use in CI, and I think we'd usually use a `# noqa:<something>` comment on the prevoius line.

andy-slac: Indeed, it's a copy&paste from CmdLineTask and I did not give any thought to it. Will fix it with `noqa:F401`.

PR #41, opened 2020-01-17
DM-22301: Improve printing of task configs

andy-slac:
The `--show=config` option now prints configuration without imports. New
option `--show=dump-config[=Task]` now dumps complete configuration for
a task (or all tasks) with all imports.

Config history option was improved as well, one can use natural configuration
syntax for configurable fields, exactly as it appears in config overrides, e.g.
`--show history="measurePsf.psfDeterminer['pca'].lam"`.

### DIFF
--- a/python/lsst/ctrl/mpexec/cmdLineParser.py
+++ b/python/lsst/ctrl/mpexec/cmdLineParser.py
@@ -459,11 +459,14 @@ def makeParser(fromfile_prefix_chars='@', parser_class=ArgumentParser, **kwargs)
 
         subparser.add_argument("--show", metavar="ITEM|ITEM=VALUE", action="append", default=[],
                                help="Dump various info to standard output. Possible items are: "
-                               "`config', `config=[Task/]<PATTERN>' or "
+                               "`config', `config=[Task::]<PATTERN>' or "
                                "`config=[Task::]<PATTERN>:NOIGNORECASE' to dump configuration "
-                               "possibly matching given pattern; `history=<FIELD>' to dump "
-                               "configuration history for a field,  field name is specified as "
-                               "[Task::][SubTask.]Field;  `pipeline' to show pipeline composition; "
+                               "fields possibly matching given pattern and/or task label; "
+                               "`history=<FIELD>' to dump configuration history for a field, "
+                               "field name is specified as [Task::][SubTask.]Field; "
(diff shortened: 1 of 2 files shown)


python/lsst/ctrl/mpexec/cmdLineParser.py
```diff
@@ -459,11 +459,14 @@ def makeParser(fromfile_prefix_chars='@', parser_class=ArgumentParser, **kwargs)
...
                                "`config=[Task::]<PATTERN>:NOIGNORECASE' to dump configuration "
-                               "possibly matching given pattern; `history=<FIELD>' to dump "
-                               "configuration history for a field,  field name is specified as "
-                               "[Task::][SubTask.]Field;  `pipeline' to show pipeline composition; "
+                               "fields possibly matching given pattern and/or task label; "
+                               "`history=<FIELD>' to dump configuration history for a field, "
+                               "field name is specified as [Task::][SubTask.]Field; "
+                               "`dump-config', `dump-config=Task' to dump complete configuration "
+                               "for a task given its label or all tasks; "
+                               "`pipeline' to show pipeline composition; "
```
TallJimbo: If you're intentionally adding two spaces only after semicolons, it'd be a lot clearer to read the code if you consistently put all spaces at the end of lines (same for unchanged lines below).  It took a bit of staring to realize that the number of spaces between words here might have been intentional.

andy-slac: Hmm, I think it's a single space after semicolons, it was not very consistent before but I tried to get rid of all double spaces. Are you confused by a backtick after double quote or does you browser display space for backicks? 🙂

TallJimbo: I was referring to the fact that you've got some spaces at the beginnings of lines and some spaces and the end of lines, and sometimes that concatenates into two spaces.

andy-slac: I think in this particular case everything is consistent, all spaces are only at the end of strings (unless we are looking at different things somehow).

TallJimbo: Ha!  I was indeed confused by backticks, which is crazy because when I was looking for backticks, I saw them, and yet when I wasn't specifically looking for them my eyes saw them as spaces.  Sorry for the noise.

PR #47, opened 2020-04-23
DM-24370: Support extensible scheduling in pipetask

andy-slac:
Executor classes have been refactored to better separate concerns. An
optional "graph fixup" instance can be passed to graph executor which
allows some modifications of the execution graph (e.g. changing
dependencies between quanta). Command line now has an option to import
and instantiate graph fixup object and pass it to executor. Unit test
was added to verify new functionality.

### DIFF
--- a/python/lsst/ctrl/mpexec/execFixupDataId.py
+++ b/python/lsst/ctrl/mpexec/execFixupDataId.py
@@ -1,0 +1,111 @@
+# This file is part of ctrl_mpexec.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (http://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+__all__ = ['ExecutionGraphFixup']
+
+from typing import Any, Iterable, Sequence, Tuple, Union
+
+from lsst.pipe.base import QuantumIterData
+from .executionGraphFixup import ExecutionGraphFixup
+
+
+class ExecFixupDataId(ExecutionGraphFixup):
+    """Implementation of ExecutionGraphFixup for ordering of tasks based
+    on DataId values.
+
+    This class is a trivial implementation mostly useful as an example,
+    though it can be used to make actual fixup instances by defining
+    a method that instantiates it, e.g.::
+
+        # lsst/ap/verify/ci_fixup.py
+
+        from lsst.ctrl.mpexec.execFixupDataId import ExecFixupDataId
+
+        def assoc_fixup():
+            return ExecFixupDataId(taskLabel="ap_assoc",
+                                   dimensions=("visit", "detector"))
+
+
+    and then executing pipetask::
+
+        pipetask run --graph-fixup=lsst.ap.verify.ci_fixup.assoc_fixup ...
+
+    This will add new dependencies between quanta executed by the task with
+    label "ap_assoc". Quanta with higher visit number will depend on quanta
+    with lower visit number and their execution will wait until lower visit
+    number finishes.
+
+    Parameters
+    ----------
+    taskLabel : `str`
+        The label of the task for which to add dependencies.
+    dimensions : `str` or sequence [`str`]
+        One or more dimension names, quanta execution will be ordered
+        according to values of these dimensions.
+    reverse : `bool`, optional
+        If `False` (default) then quanta with higher values of dimensions
+        will be executed after quanta with lower values, otherwise the order
+        is reversed.
+    """
+
+    def __init__(self, taskLabel: str, dimensions: Union[str, Sequence[str]], reverse: bool = False):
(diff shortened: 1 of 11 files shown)


python/lsst/ctrl/mpexec/execFixupDataId.py
```diff
@@ -0,0 +1,103 @@
...
+from .executionGraphFixup import ExecutionGraphFixup
+
+
+class ExecFixupDataId(ExecutionGraphFixup):
+    """Implementation of ExecutionGraphFixup for ordering of tasks based
+    on DataId values.
+
+    This class is a trivial implementation mostly useful as an example,
+    though it can be used to make actual fixup instances by defining
+    a method that instanciates it, e.g.::
```
timj, line 36: typo: instantiates

python/lsst/ctrl/mpexec/execFixupDataId.py
```diff
@@ -0,0 +1,103 @@
...
+        def assoc_fixup():
+            return ExecFixupDataId(taskLabel="ap_assoc",
+                                   dimensions=("visit", "detector"))
+
+
+    and then executing pipetask::
+
+        pipetask run --graph-fixup=lsst.ap.verify.ci_fixup.assoc_fixup ...
+
+    This will all new dependencies between quanta executed by the task with
```
timj, line 51: Typo "this will all new"

python/lsst/ctrl/mpexec/execFixupDataId.py
```diff
@@ -0,0 +1,103 @@
...
+    This will add new dependencies between quanta executed by the task with
+    label "ap_assoc". Quanta with higher visit number will depend on quanta
+    with lower visit number and their execution will wait until lower visit
+    number finishes.
+
+    Parameters
+    ----------
+    taskLabel : `str`
+        The label of the task for which to add dependencies.
+    dimensions : `str` or iterable of `str`
```
kfindeisen, line 60: If I understand the code right, this should be "sequence of `str`". The sort order would be ambiguous if you passed in, for example, a `set`.

python/lsst/ctrl/mpexec/execFixupDataId.py
```diff
@@ -0,0 +1,103 @@
...
+        quanta = list(quanta)
+        taskQuanta = [q for q in quanta if q.taskDef.label == self.taskLabel]
+        if not taskQuanta:
+            raise ValueError(f"Cannot find task with label {self.taskLabel}")
+        # order quanta by dataId dimensions
+        taskQuanta.sort(key=self._key, reverse=self.reverse)
+        for i, qdata in enumerate(taskQuanta):
+            # add previous quanta as dependcency
+            if i > 0:
+                qdata.dependencies |= frozenset([taskQuanta[i-1].index])
```
kfindeisen, line 102: Hmm... won't this mean that even if I sort by `visit` only, all quanta for different detectors but the same visit will still be executed sequentially? That should be fine for current testing (association is a negligible part of our CI-scale running time), but may slow things down as runs get bigger.

python/lsst/ctrl/mpexec/execFixupDataId.py
kfindeisen: Preferred syntax [uses bracket notation](https://developer.lsst.io/python/numpydoc.html#py-docstring-parameter-types-sequences).

python/lsst/ctrl/mpexec/executionGraphFixup.py
```diff
@@ -0,0 +1,65 @@
...
+        """Update quanta in a graph.
+
+        Potentially anything in the graph could be changed if it does not
+        break executor assumptions. Returned quanta will be re-ordered by
+        executor, if modifications result in a dependency cycle the executor
+        will raise an exception.
+
+        Parameters
+        ----------
+        quanta : iterable of `~lsst.pipe.base.QuantumIterData`
```
kfindeisen, line 57:
```suggestion
        quanta : iterable [`~lsst.pipe.base.QuantumIterData`]
```

python/lsst/ctrl/mpexec/execFixupDataId.py
andy-slac: Indeed, the code as it is written now won't even work with iterable. I'm inclined to change the arguments to positionals:

    def __init__(self, taskLabel: str, *dimensions: str, *, reverse: bool = False)

to avoid that ordering ambiguity. Do you like this better?

Regarding bracket notation - I'm not quite sure that it applies here, strictly speaking iterable is not a container, and I did not find an example in the developer guide for that, that is why I decided to use `iterable of`

python/lsst/ctrl/mpexec/execFixupDataId.py
andy-slac: This is true, I think I can update this implementation to work with multiple quanta per key, hang on.

python/lsst/ctrl/mpexec/executionGraphFixup.py
andy-slac: Fine, I'll change it 🙂

python/lsst/ctrl/mpexec/execFixupDataId.py
kfindeisen: I don't understand why you would want to use variadic arguments here, they seem more error-prone. Why can't you simply document the argument as a sequence rather than any iterable?

andy-slac: I like variadic better because it fixes the order and most of the time dimensions will be known in advance so in my view it's more natural to write:

    Class("label", "visit", "detector")

than 

    Class("label", ("visit", "detector"))

Still if you prefer single argument I'll do it that way.

python/lsst/ctrl/mpexec/execFixupDataId.py
andy-slac: @kfindeisen, I have updated the algorithm, I checked that it works but please review it to see that it also works for you.

kfindeisen: No objections 👍

PR #51, opened 2020-05-13
DM-24797: Store per-run information (configs, software versions) in butler repo

andy-slac:
PreExecInit now saves task configurations for every task in a pipeline,
the names of dataset types are fixed as `{taskLabel}_config`. It also saves
package versions in a dataset with fixed type name "packages". Dataset
types are registered together with all other dataset types when 
--register-dataset-types option is specified.

### DIFF
--- a/python/lsst/ctrl/mpexec/preExecInit.py
+++ b/python/lsst/ctrl/mpexec/preExecInit.py
@@ -30,6 +30,7 @@ import itertools
 # -----------------------------
 #  Imports for other modules --
 # -----------------------------
+from lsst.daf.butler import DatasetType
 from lsst.pipe.base import PipelineDatasetTypes
 
 _LOG = logging.getLogger(__name__.partition(".")[2])
@@ -81,6 +82,7 @@ class PreExecInit:
         # is consistent with what tasks would save
         if saveInitOutputs:
             self.saveInitOutputs(graph)
+            self.saveConfigs(graph)
 
     def initializeDatasetTypes(self, graph, registerDatasetTypes=False):
         """Save or check DatasetTypes output by the tasks in a graph.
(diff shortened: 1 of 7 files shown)


python/lsst/ctrl/mpexec/preExecInit.py
```diff
@@ -172,3 +182,42 @@ def saveInitOutputs(self, graph):
...
+        ----------
+        graph : `~lsst.pipe.base.QuantumGraph`
+            Execution graph.
+
+        Raises
+        ------
+        Exception
+            Raised if ``skipExisting`` is `False` and datasets already
+            exists. Content of a butler collection may be changed if
+            exception is raised.
```
TallJimbo, line 212: We could make this exception safe by wrapping it in `with self.butler.transaction()`, I think.

python/lsst/ctrl/mpexec/preExecInit.py
```diff
@@ -221,3 +233,36 @@ def logConfigMismatch(msg):
...
+            Raised if ``checkExisting`` is ``True`` but versions are not
+            compatible.
+        """
+        packages = Packages.fromSystem()
+        datasetType = "packages"
+        oldPackages = self.butler.get(datasetType, {}) if self.skipExisting else None
+        if oldPackages is not None:
+            # Note that because we can only detect python modules that have been imported, the stored
+            # list of products may be more or less complete than what we have now.  What's important is
+            # that the products that are in common have the same version.
```
TallJimbo, line 257: I think the set of imported packages will be sufficiently complete after we've loaded all `PipelineTask` config instances, and probably not complete before that.  Do you happen to know if the sequencing of operations is such that this will be called after that happens?

python/lsst/ctrl/mpexec/preExecInit.py
andy-slac: Indeed, let me try that.

python/lsst/ctrl/mpexec/preExecInit.py
andy-slac: I though about that and figured that it should work because the complete graph is in memory (with some caveat) and it means that config instances are loaded too which should import all relevant code.
The caveat is that BPS is executing this not for the whole graph but one task at a time so there is only one task configuration loaded and not whole graph. But this still should work because we update versions incrementally if they don't conflict. And I think BPS will probably do --init-only on the whole graph in the future.

python/lsst/ctrl/mpexec/preExecInit.py
andy-slac: That works fine with ci_hsc_gen3, will add that.

PR #54, opened 2020-05-22
DM-25016: Fix reading of QuantumGraph from pickle file

andy-slac:
When reading QuantumGraph from pickle DimensionUniverse instance must be
initialized, this fix makes registry instance before un-pickling which
creates all necessary instances. This trivial fix exposed an issue with
my dumb unit test that also needed proper fix.

### DIFF
--- a/tests/test_cmdLineFwk.py
+++ b/tests/test_cmdLineFwk.py
@@ -33,7 +33,8 @@ import unittest
 from lsst.ctrl.mpexec.cmdLineFwk import CmdLineFwk
 from lsst.ctrl.mpexec.cmdLineParser import (_ACTION_ADD_TASK, _ACTION_CONFIG,
                                             _ACTION_CONFIG_FILE, _ACTION_ADD_INSTRUMENT)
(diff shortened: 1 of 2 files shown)


tests/test_cmdLineFwk.py
```diff
@@ -66,6 +67,23 @@ def makeTmpFile(contents=None):
...
+def makeSQLiteRegistry():
+    """Context manager to create new empty registry database.
+
+    Yields
+    ------
+    config : `RegistryConfig`
+        Registry configuration for initialized registry database.
+    """
+    with makeTmpFile() as filename:
+        uri = f"sqlite:///{filename}"
```
timj, line 80: Does this have to be a file? Can't it be an in-memory sqlite?

andy-slac: In-memory does not work in this case, there is a discussion on JIRA about this issue. Basically if you want to open registry in read-only mode then you can't use in-memory because every session for in-memory sqlite makes it's own database which cannot be shared, and read-only database cannot be initialized with all the stuff that we need.

PR #55, opened 2020-06-03
DM-25222: Fix handling of skip-existing in PreExecInit

andy-slac:
PreExecInit expected butler.get() to return None for missing datasets,
while in reality it raises LookupError. Fixing this and also adding unit
test for this.

### DIFF
--- a/python/lsst/ctrl/mpexec/preExecInit.py
+++ b/python/lsst/ctrl/mpexec/preExecInit.py
@@ -182,14 +182,16 @@ class PreExecInit:
                     # check if it is there already
                     _LOG.debug("Retrieving InitOutputs for task=%s key=%s dsTypeName=%s",
                                task, name, attribute.name)
-                    objFromStore = self.butler.get(attribute.name, {})
-                    if objFromStore is not None:
+                    try:
+                        objFromStore = self.butler.get(attribute.name, {})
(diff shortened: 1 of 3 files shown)


python/lsst/ctrl/mpexec/preExecInit.py
```diff
@@ -182,14 +182,16 @@ def saveInitOutputs(self, graph):
                     # check if it is there already
                     _LOG.debug("Retrieving InitOutputs for task=%s key=%s dsTypeName=%s",
                                task, name, attribute.name)
-                    objFromStore = self.butler.get(attribute.name, {})
-                    if objFromStore is not None:
+                    try:
+                        objFromStore = self.butler.get(attribute.name, {})
                         # Types are supposed to be identical.
```
timj, line 187: Datastore.get does check that the returned python object is an instance of the StorageClass python type. It does not require that types match directly. Your test here is more stringent than butler -- does it ever trigger?

andy-slac: I have no idea, probably not because we did not see any questions/complaints about this. What we want to do is to make sure that objects are identical but there is no way to do it now for arbitrary types, checking that types are identical is the minimum that we can do here.

PR #61, opened 2020-07-24
DM-26088: Skip lsst.log reset if LSST_LOG_CONFIG is set

hsinfang:


python/lsst/ctrl/mpexec/cmdLineFwk.py
```diff
@@ -511,7 +512,8 @@ def configLog(longlog, logLevels):
             message_fmt = "%c %p: %m%n"
 
         # global logging config
-        lsst.log.configure_prop(_LOG_PROP.format(message_fmt))
+        if ("LSST_LOG_CONFIG" not in os.environ) or (not os.path.isfile(os.environ["LSST_LOG_CONFIG"])):
```
andy-slac, line 515: Could probably shorten this to:

    if not os.path.isfile(os.environ.get("LSST_LOG_CONFIG", "")):

PR #62, opened 2020-07-31
DM-26181: Ensure that filters are defined in pipetask

andy-slac:
Before executing a task scan its dataIds and see if they have an
instrument name, if so then make instrument instance which initializes
filter singleton with all instrument filters.

### DIFF
--- a/python/lsst/ctrl/mpexec/singleQuantumExecutor.py
+++ b/python/lsst/ctrl/mpexec/singleQuantumExecutor.py
@@ -32,6 +32,7 @@ from itertools import chain
 # -----------------------------
 from .quantumGraphExecutor import QuantumExecutor
 from lsst.log import Log
+from lsst.obs.base import Instrument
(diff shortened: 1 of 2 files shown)


python/lsst/ctrl/mpexec/singleQuantumExecutor.py
```diff
@@ -233,3 +238,35 @@ def runQuantum(self, task, quantum, taskDef, butler):
...
+        This will need revision when filter singleton disappears.
+        """
+        oneInstrument = None
+        for datasetRefs in chain(quantum.predictedInputs.values(), quantum.outputs.values()):
+            for datasetRef in datasetRefs:
+                dataId = datasetRef.dataId
+                instrument = dataId.get("instrument")
+                if instrument is not None:
+                    if oneInstrument is not None:
+                        assert instrument == oneInstrument
```
timj, line 269: Can you add a message to this along the lines of "Currently require that only one instrument is used per graph" or something?

PR #65, opened 2020-08-11
DM-25627 make qgraph and run subcommands for pipetask

n8pease:


python/lsst/ctrl/mpexec/cli/script/qgraph.py
```diff
@@ -0,0 +1,156 @@
...
+        Format string of locations for storing individual quantum graph
+        definition (pickle files). The curly brace {} in the input string will
+        be replaced by a quantum number.
+    qgraph_dot : `str` or `None`
+        Path location for storing GraphViz DOT representation of a quantum graph.
+    butler_config : `str`, `dict`, or `lsst.daf.butler.Config`
+        If `str`, `butler_config` is the path location of the gen3
+        butler/registry config file. If `dict`, `butler_config` is key value
+        pairs used to init or update the `lsst.daf.butler.Config` instance. If
+        `Config`, it is the object used to configure a Butler.
```
andy-slac, line 68: Do we need all these options in CLI?

python/lsst/ctrl/mpexec/cli/script/qgraph.py
```diff
@@ -0,0 +1,156 @@
...
+        replaced run unless `prune-replaced` is also True. Requires `output`,
+        and `extend_run` must be `None`.
+    prune_replaced : "unstore", "purge", or `None`.
+        If not `None`, delete the datasets in the collection replaced by
+        `replace_run`, either just from the datastore ("unstore") or by
+        removing them and the RUN completely ("purge"). Requires `replace_run`.
+    data_query : `str`
+        User query selection expression.
+    show : `list` [`str`] or `None`
+        Descriptions of what to dump to stdout.
```
andy-slac, line 103: There is a lot of duplication between this docstring and help strings in option definitions. I'm not sure that we need to repeat them here, this is not a public API, and we want to avoid repetition for this sort of descriptions.

python/lsst/ctrl/mpexec/cli/script/qgraph.py
```diff
@@ -0,0 +1,156 @@
...
+    #         "Run 'build' command before 'qgraph' or provide a serialzed qgraph file location.")
+
+    if log_level is not None:
+        CliLog.setLogLevels(log_level)
+
+    class MakeGraphArgs:
+        """A container class for arguments to CmdLineFwk.makeGraph, whose
+        API (currently) is written to accept inputs from argparse in a generic
+        container class.
+        """
```
andy-slac, line 124: I think `types.SimpleNamespace` class would work better for this: https://docs.python.org/3/library/types.html#types.SimpleNamespace

python/lsst/ctrl/mpexec/cli/script/run.py
```diff
@@ -0,0 +1,190 @@
...
+    args = RunArgs(do_raise=do_raise, graph_fixup=graph_fixup, init_only=init_only, no_versions=no_versions,
+                   processes=processes, profile=profile, skip_init_writes=skip_init_writes, timeout=timeout,
+                   register_dataset_types=register_dataset_types, butler_config=butler_config, input=input,
+                   output=output, output_run=output_run, extend_run=extend_run, replace_run=replace_run,
+                   prune_replaced=prune_replaced, data_query=data_query, skip_existing=skip_existing,
+                   debug=debug)
+
+    f = CmdLineFwk()
+    taskFactory = TaskFactory()
+    f.runPipeline(qgraph, taskFactory, args)
```
andy-slac, line 190: Same comments as for qgraph.py

python/lsst/ctrl/mpexec/cli/opt/input.py
```diff
@@ -0,0 +1,61 @@
...
+        super().__init__(forward=forward)
+        self.help = help or self.defaultHelp()
+        self.metavar = metavar
+        self.multiple = multiple
+        self.required = required
+
+    def __call__(self, f):
+        return click.option(*self.optionFlags(), cls=MWOption,
+                            callback=self.callback,
+                            default=None if self.required else list(),
```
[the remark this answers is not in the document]
n8pease: just a general preference for immutable objects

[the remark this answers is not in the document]
timj: If you want immutable shouldn't it be a tuple?

n8pease:
@andy-slac 

> Looks OK, check few comments. It feels like we are overdoing it with a module per command line option. There is so much boilerplate and duplication everywhere it's hard to read it.

It's a fair point. I was hoping to promote reuse by just defining a shared option for each. It may err on the side of too much boilerplate. I might be able to make the shared option definition more concise, I'll think on that.

python/lsst/ctrl/mpexec/cli/opt/input.py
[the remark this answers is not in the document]
n8pease: hah. Yes. 
> Is list() better than []
Trying again: I don't trust that passing `default=[]` won't end up sharing a list instance among multiple instances of the Option, and depending on how downstream code uses the default value chaos may ensue. `list()` is guaranteed to create a new list instance each time it is used and is safer. OTOH, `()` or `tuple()` would both be fine since `()` is immutable. Probably `()` is slightly better because as I understand it `tuple()` results in an extra line of code being executed (`return ()` or similar)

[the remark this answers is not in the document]
n8pease: and, click won't take a tuple as the default value of an option, so I'm going to stick with `list()`

python/lsst/ctrl/mpexec/cli/script/qgraph.py
n8pease: no, and CLI says it must be a file. But at this level (according to downstream documentation), you could pass in any of these things, and CLI could be extended to accept e.g. kv pairs.

python/lsst/ctrl/mpexec/cli/script/qgraph.py
n8pease: I really don't know how to slice this in python. I think it's super maddening to look at a function without any documentation and have to reverse engineer the types and intents of each argument, or to know where to look up or down the stack (after having figured out what the stack will be) to find the documentation and to then know if that documentation applies here. Yes this is verbose, but at least when I'm here I know what was intended by the author or last person to make changes and IMO that's better than the guessing game.

python/lsst/ctrl/mpexec/cli/script/qgraph.py
n8pease: from the docs:
> SimpleNamespace may be useful as a replacement for class NS: pass

I had basically done that (`class NS: pass`) for the `build` subcommand, @timj did not like it and that's why I wrote the Args classes.

timj: What didn't I like?

n8pease: in https://github.com/lsst/ctrl_mpexec/pull/58, search for "cute" (I don't see a way to link to the particular conversation)

n8pease: `Args` was declared 
```
class Args
    pass
args = Args()
args.qgraph = qgraph
args.save_qgraph = save_qgraph
<etc>
```

timj: I think you are over interpreting my comment. My final quote was:

> At the very least a comment like you just made in the class description for Args would be useful.

(I haven't looked at the current context -- SimpleNamespace does look useful as an explicit declaration of intent)

python/lsst/ctrl/mpexec/cli/opt/input.py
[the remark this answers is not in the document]
andy-slac: My question was mostly rhetorical, there is no real difference between `list()` and `[]`, both of them make a new empty list (but latter is shorter). What you pass as a parameter is not going to be shared, because it's a new list on every call. The sharing argument is only valid for the default argument value in the declaration of a method, there indeed you should not use mutable values as defaults.

python/lsst/ctrl/mpexec/cli/script/qgraph.py
andy-slac: My concern is that description here and description of CLI arguments can get out of sync. We probably need `SuperClick` which can read method docstring and build arguments from that 🙂

n8pease:
Thinking about ways I could reduce the size of the shared option declarations I've realized I can use `functools.partial` to bind option decorator arguments to reduce almost all the boilerplate that was needed for `MWOptionDecorator` subclasses, and can instead have `MWOptionDecorator` instances be the option decorators that are used. (@andy-slac this is very similar but slightly more than what I presented to you in Slack). `MWOptionDecorator` becomes like this
```
class MWOptionDecorator:

    def __init__(self, *param_decls, **kwargs):
        self.partialOpt = partial(click.option, *param_decls, cls=MWOption, **kwargs)
        opt = click.Option(param_decls, **kwargs)
        self._name = opt.name
        self._opts = opt.opts

    def name(self):
        return self._name

    def opts(self):
        return self._opts

    def __call__(self, *args, **kwargs):
        return self.partialOpt(*args, **kwargs)
```
and its use to define shared options is very terse, for example:
```
butler_config_option = MWOptionDecorator("-b", "--butler-config",
                                         help="Location of the gen3 butler/registry config file.",
                                         type = MWPath(dir_okay=False, file_okay=True, readable=True))

data_query_option = MWOptionDecorator("-d", "--data-query",
                                      help="User data selection expression."
                                      metavar="QUERY")
```
I think this will be a big improvement.

PR #66, opened 2020-08-20
DM-26136: Improve handling of crashes in pipetask

andy-slac:
Replaced use of multiprocessing Pool with manual process management based on `Process` class. Crashes in sub-processes should be handled correctly now. Exceptions are not propagated to main process, they will appear in the log when subprocess fail, main process will just list all tasks that failed.

Implemented "keep running" as default option, there is new command line option `--fail-fast` which turns on old behavior.

### DIFF
--- a/python/lsst/ctrl/mpexec/mpGraphExecutor.py
+++ b/python/lsst/ctrl/mpexec/mpGraphExecutor.py
@@ -26,6 +26,7 @@ __all__ = ["MPGraphExecutor", "MPGraphExecutorError", "MPTimeoutError"]
 # -------------------------------
 import logging
 import multiprocessing
+import time
 
 # -----------------------------
 #  Imports for other modules --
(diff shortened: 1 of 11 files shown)


python/lsst/ctrl/mpexec/mpGraphExecutor.py
```diff
@@ -157,74 +293,85 @@ def _executeQuantaMP(self, iterable, butler):
...
+                    # all dependencies have completed, can start new job
+                    if len(jobs.running()) < self.numProc:
+                        _LOG.debug("Sumbitting %s", job)
+                        job.start(butler, self.quantumExecutor)
+
+            # Here we want to wait until one of the running jobs completes
+            # but multiprocessing does not provide an API for that, for now
+            # just sleep a little bit and go back to the loop.
+            if jobs.running():
+                time.sleep(0.1)
```
timj, line 364: I can't help pondering asyncio in cases like this where we are sleeping and looping.

andy-slac: I'm not very proud of this two lines of code, it's stupid and inefficient. We could certainly do fancier things here, in the context of pipetask it probably does not matter very much, but generally I agree we want to make things behaving better. Maybe something for later ticket together with finally getting rid of `multiprocessing` stuff.

PR #67, opened 2020-08-22
DM-26415: remove chained commands from pipetask2 and DM-26388: pipetask2 --butler-config should accept a repository URI

n8pease:


python/lsst/ctrl/mpexec/cli/opt/optionGroups.py
```diff
@@ -0,0 +1,123 @@
...
+from lsst.daf.butler.cli.utils import option_section, split_kv, unwrap
+import lsst.obs.base.cli.opt as obsBaseOpts
+import lsst.daf.butler.cli.opt as dafButlerOpts
+from . import options as ctrlMpExecOpts
+
+instrumentOptionHelp = ("Add an instrument which will be used to load config overrides when defining a "
+                        "pipeline. This must be the fully qualified class name.")
+
+
+class OptionGroup():
```
andy-slac, line 38: Did you want to add base class inside parentheses?

python/lsst/ctrl/mpexec/cli/opt/optionGroups.py
```diff
@@ -0,0 +1,123 @@
...
+    """Base class for an option group decorator. Requires the option group
+    subclass to have a property called `decorator`."""
+
+    def __call__(self, f):
+        for decorator in reversed(self.decorators):
+            f = decorator(f)
+        return f
+
+
+class pipeline_build_options(OptionGroup):  # noqa: N801
```
andy-slac, line 48: Short one-line docstring could be useful for these classes.

python/lsst/ctrl/mpexec/cli/opt/optionGroups.py
```diff
@@ -0,0 +1,123 @@
...
+
+class qgraph_options(OptionGroup):  # noqa: N801
+
+    def __init__(self):
+        self.decorators = [option_section(sectionText="Quantum graph building options:"),
+                           ctrlMpExecOpts.qgraph_option(),
+                           ctrlMpExecOpts.skip_existing_option(),
+                           ctrlMpExecOpts.save_qgraph_option(),
+                           ctrlMpExecOpts.save_single_quanta_option(),
+                           ctrlMpExecOpts.qgraph_dot_option()]
```
andy-slac, line 75: Nitpick - there is a mixture of wrapping/indentation styles in different classes like

    self.decorators = [decorator1,
                       decorator2]

vs.

    self.decorators = [
        decorator1,
        decorator2
    ]

I personally prefer latter style, feels more readable as it lets you pack longer lines without more wrapping.

python/lsst/ctrl/mpexec/cli/script/build.py
```diff
@@ -50,6 +52,10 @@ def build(order_pipeline=None, pipeline=None, pipeline_actions=(), pipeline_dot=
         Path location for storing resulting pipeline definition in YAML format.
     show : `list` [`str`]
         Descriptions of what to dump to stdout.
+    kwargs : `dict` [`str`, `str`]
+        Ignored; click commands may accept options for more than one script
+        function and pass all the option kwargs to each of the script functions
+        which ingore these unused kwargs.
```
andy-slac, line 58: Hmm, this is potentially problematic, if for example client misspells option/keyword name, you won't catch it here. Could we do something safer here (and other places like this)?

n8pease: @andy-slac I think there is not a good reason for the explicit kwargs to have default values. What if I removed them? That way if any of them are misspelled, and therefore missing, there will be an error.

n8pease: (Click generates default value for all the options, the command function gets called with a kwarg for each of the `@option`s it is decorated with.

python/lsst/ctrl/mpexec/cli/opt/optionGroups.py
n8pease: No, but I guess I can remove the parens?

python/lsst/ctrl/mpexec/cli/opt/optionGroups.py
n8pease: Oh, latter is nice. 👍

python/lsst/ctrl/mpexec/cli/script/build.py
andy-slac: Having default values here and in click indeed bothers me too. If we say that all arguments are required than indeed we can drop defaults here and that should reduce the probability for spelling errors.

python/lsst/ctrl/mpexec/cli/opt/optionGroups.py
andy-slac: 👍

PR #69, opened 2020-08-31
Fix Postgres connection error in multi-process case (DM-26539)

andy-slac:
My previous workaround for passing butler to subprocess was broken
because new connection was made in the parent process. Replacing with a
new workaround that passes butler in pickle form across fork.

### DIFF
--- a/python/lsst/ctrl/mpexec/mpGraphExecutor.py
+++ b/python/lsst/ctrl/mpexec/mpGraphExecutor.py
@@ -24,10 +24,10 @@ __all__ = ["MPGraphExecutor", "MPGraphExecutorError", "MPTimeoutError"]
 # -------------------------------
 #  Imports of standard modules --
 # -------------------------------
-import copy
 from enum import Enum
 import logging
 import multiprocessing
+import pickle
 import time
 
 # -----------------------------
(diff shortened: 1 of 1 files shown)


python/lsst/ctrl/mpexec/mpGraphExecutor.py
```diff
@@ -74,19 +74,37 @@ def start(self, butler, quantumExecutor):
             Executor for single quantum.
         """
         # Butler can have live database connections which is a problem with
-        # fork-type activation. Make a copy of butler, this guarantees that
-        # no database is open right after copy.
-        butler = copy.copy(butler)
+        # fork-type activation. Make a pickle of butler to pass that actross
```
timj, line 77: type: actross

andy-slac:
> 
> 
> Is this change because pickling guarantees that each fork gets a new butler, whereas before every fork was getting the same butler?

Not quite. Before this each fork also had a new butler but that instance was shared with parent process (butler was instantiated by parent) and parent and child had a connection to database which parent closed immediately after "passing" butler to child process. That clearly confused server because child tried to use connection that was closed already (whatever that means for Postgres server). New code avoids instantiating butler in parent process, butler is created by child process now, parent and child only share pickle bytes which has no state.

PR #71, opened 2020-09-11
DM-26371 Update code to support new QuantumGraph type

natelust:


python/lsst/ctrl/mpexec/execFixupDataId.py
```diff
@@ -60,52 +60,24 @@ def assoc_fixup():
...
-    reverse : `bool`, optional
-        If `False` (default) then quanta with higher values of dimensions
-        will be executed after quanta with lower values, otherwise the order
-        is reversed.
     """
 
     def __init__(self, taskLabel: str, dimensions: Union[str, Sequence[str]], reverse: bool = False):
         self.taskLabel = taskLabel
         self.dimensions = dimensions
-        self.reverse = reverse
```
andy-slac, line 72: You did not remove it from parameter list? Why `reverse` is not supported, is it impossible in a new world?

python/lsst/ctrl/mpexec/execFixupDataId.py
```diff
@@ -60,52 +60,24 @@ def assoc_fixup():
...
             raise ValueError(f"Cannot find task with label {self.taskLabel}")
-        # order keys
-        keys = sorted(keyQuanta.keys(), reverse=self.reverse)
-        # for each quanta in a key add dependency to all quanta in a preceding key
-        for prev_key, key in zip(keys, keys[1:]):
-            prev_indices = frozenset(qdata.index for qdata in keyQuanta[prev_key])
-            for qdata in keyQuanta[key]:
-                qdata.dependencies |= prev_indices
-        return quanta
+        quanta = list(graph.quantaForTask(taskDef))
```
andy-slac, line 77: What order do they come out? Previous version ordered them by dataId, can't we do the same here?

python/lsst/ctrl/mpexec/mpGraphExecutor.py
```diff
@@ -54,14 +56,16 @@ class _Job:
 
     Parameters
     ----------
-    qdata : `~lsst.pipe.base.QuantumIterData`
+    qdata : `~lsst.pipe.base.QuantumNode`
```
andy-slac, line 59: I would rename `qdata` to `qnode`.

python/lsst/ctrl/mpexec/mpGraphExecutor.py
```diff
@@ -54,14 +56,16 @@ class _Job:
...
         Quantum and some associated information.
     """
```
andy-slac, line 61: CAn you add docstring for two new arguments.

python/lsst/ctrl/mpexec/mpGraphExecutor.py
```diff
@@ -164,7 +168,7 @@ def finishedIds(self):
         jobsIds : `set` [`int`]
             Set of integer job IDs.
         """
-        return set(job.qdata.index for job in self.jobs if job.state == JobState.FINISHED)
+        return set(job.qdata for job in self.jobs if job.state == JobState.FINISHED)
```
andy-slac, line 171: Should it be `job.index` instead? Same for two other changes below. If it needs to return nodes instead of index then it's better to rename methods and update docstrings.

python/lsst/ctrl/mpexec/mpGraphExecutor.py
```diff
@@ -227,24 +231,24 @@ def __init__(self, numProc, timeout, quantumExecutor, *, failFast=False, executi
...
-            Quanta as originated from a quantum graph.
+        graph : `QuantumGraph`
+            `QuantumGraph` to modify
 
         Returns
         -------
-        quantaIter : iterable of `~lsst.pipe.base.QuantumIterData`
-            Possibly updated set of quanta, properly ordered for execution.
+        graph : `QuantumGraph`
+            Modified `QuantumGraph`, properly ordered for execution.
```
andy-slac, line 251: I think new QuantumGRaph cannot be non-ordered, may be drop this "properly ordered..."

tests/test_executors.py
```diff
@@ -181,22 +218,18 @@ def test_mpexec_fixup(self):
 
         taskDef = TaskDefMock()
 
-        for reverse in (False, True):
```
andy-slac, line 184: Would be nice if we could keep `reverse`.

andy-slac (approved):
Looks OK on general, some renaming is probably needed in mpGraphExecutor and I would like to keep `reverse` option in FixupDataId if possible.

python/lsst/ctrl/mpexec/singleQuantumExecutor.py
```diff
@@ -210,9 +212,15 @@ def updateQuantumInputs(self, quantum, butler):
             Single Quantum instance.
         butler : `~lsst.daf.butler.Butler`
             Data butler.
+
+        Returns
+        -------
+        update : `~lsst.daf.butler.Quantum`
```
andy-slac, line 218: Can you call it `quantum` instead of `update`? Or `updatedQuantum`?

andy-slac (approved):
Looks good.

python/lsst/ctrl/mpexec/mpGraphExecutor.py
natelust: They were not needed any more and were removed instead

python/lsst/ctrl/mpexec/execFixupDataId.py
natelust: Added back in

python/lsst/ctrl/mpexec/execFixupDataId.py
natelust: Added back in

PR #72, opened 2020-09-11
DM-24414: Implement support for --prune-replaced option

andy-slac:
It is few lines of code change but needed lots of updates in test
harness as testing this feature needs persistent registry and datastore.

### DIFF
--- a/python/lsst/ctrl/mpexec/cmdLineFwk.py
+++ b/python/lsst/ctrl/mpexec/cmdLineFwk.py
@@ -213,7 +213,7 @@ class _ButlerFactory:
             self.outputRun = None
         else:
             raise ValueError("Cannot write without at least one of (--output, --output-run).")
-        self.inputs = list(CollectionSearch.fromExpression(args.input))
+        self.inputs = list(CollectionSearch.fromExpression(args.input)) if args.input else []
 
     def check(self, args: argparse.Namespace):
         """Check command-line options for consistency with each other and the
@@ -353,11 +353,21 @@ class _ButlerFactory:
             chainDefinition = list(self.output.chain if self.output.exists else self.inputs)
             if args.replace_run:
                 replaced, _ = chainDefinition.pop(0)
-                if args.prune_replaced:
-                    # TODO: DM-23671: need a butler API for pruning an
-                    # entire RUN collection, then apply it to 'replaced'
-                    # here.
-                    raise NotImplementedError("Support for --prune-replaced is not yet implemented.")
+                if args.prune_replaced == "unstore":
+                    # Remove datasets from datastore
+                    with butler.transaction():
+                        refs = butler.registry.queryDatasets(..., collections=replaced)
+                        butler.pruneDatasets(refs, unstore=True, run=replaced, disassociate=False)
+                elif args.prune_replaced == "purge":
+                    # Erase entire collection and all datasets, need to remove
+                    # collection from its chain collection first.
+                    with butler.transaction():
+                        butler.registry.setCollectionChain(self.output.name, chainDefinition)
+                        butler.pruneCollection(replaced, purge=True, unstore=True)
+                elif args.prune_replaced is not None:
+                    raise NotImplementedError(
+                        f"Unsupported --prune-replaced option '{args.prune_replaced}'."
+                    )
             chainDefinition.insert(0, self.outputRun.name)
             chainDefinition = CollectionSearch.fromExpression(chainDefinition)
             _LOG.debug("Preparing butler to write to '%s' and read from '%s'=%s",
--- a/tests/testUtil.py
+++ b/tests/testUtil.py
@@ -28,7 +28,7 @@ import itertools
 import logging
 import numpy
 
-from lsst.daf.butler import (Butler, DatasetType, CollectionSearch)
+from lsst.daf.butler import (Butler, Config, DatasetType, CollectionSearch)
 import lsst.daf.butler.tests as butlerTests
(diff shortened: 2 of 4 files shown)


python/lsst/ctrl/mpexec/cmdLineFwk.py
```diff
@@ -353,11 +353,17 @@ def makeWriteButler(cls, args: argparse.Namespace) -> Butler:
...
+                    # Remove datasets from datastore
+                    with butler.transaction():
+                        refs = butler.registry.queryDatasets(..., collections=replaced, deduplicate=True)
+                        butler.pruneDatasets(refs, unstore=True, run=replaced, disassociate=False)
+                elif args.prune_replaced == "purge":
+                    # Erase entire collection and all datasets, need to remove
+                    # collection from its chain collection first.
+                    with butler.transaction():
+                        butler.registry.setCollectionChain(self.output.name, chainDefinition)
+                        butler.pruneCollection(replaced, purge=True, unstore=True)
```
timj, line 366: can you have an else that bails out here? I know that config is supposed to restrict the options but I still feel better if we check.

tests/testUtil.py
```diff
@@ -152,7 +151,37 @@ def registerDatasetTypes(registry, pipeline):
                 registry.registerDatasetType(datasetType)
 
 
-def makeSimpleQGraph(nQuanta=5, pipeline=None, butler=None, root=None, skipExisting=False):
+def _makeButlerRepo(root, inMemory=True):
```
timj, line 154: Why did you stop using `lsst.daf.butler.tests.makeTestRepo` -- it did this configuration for you.

python/lsst/ctrl/mpexec/cmdLineFwk.py
andy-slac: I will add handling for all possible options.

tests/testUtil.py
andy-slac: makeTestRepo only makes in-memory repo, for this new feature I have to test with persistent repo because I need to instantiate multiple butlers/registries on the same data.

timj:
> It is few lines of code change but needed lots of updates in test
> 
> harness as testing this feature needs persistent registry and datastore.

You can pass a config override in so you can do what you need.

tests/testUtil.py
timj: You can pass a Config of overrides into it. For example as done in https://github.com/lsst/obs_base/blob/master/tests/test_butlerFits.py#L111  -- you can set just the registry.db part of a Config.

andy-slac: Well, I'll have to pass something for `datastore.cls` which I don't know what it has to be. Or will it work if I set it to None? Besides what I really need is a butler config, in I was thinking that `butler._config` is not public and it may be better not to use it (and I did not check that it's unmodified from what is passed originally).

timj: Do you really need butler_config? The use cases below would work if you just passed self.root wouldn't they? (since the butler would read the on-disk butler.yaml file from self.root. I'm probably missing something. There will always be a butler.yaml written into root given the options specified there (other options allow it to be written outside). I'm not going to insist on shoehorning this into makeButlerRepo though if there is a good reason not to do it.

andy-slac: OK, I re-implemented that part and also my new test code, it does not need to instantiate multiple butlers now (but needs to call some internal methods to refresh collections). I hope tests are getting better, thanks for suggestions!

python/lsst/ctrl/mpexec/cmdLineFwk.py
```diff
@@ -353,11 +353,21 @@ def makeWriteButler(cls, args: argparse.Namespace) -> Butler:
...
                 replaced, _ = chainDefinition.pop(0)
-                if args.prune_replaced:
-                    # TODO: DM-23671: need a butler API for pruning an
-                    # entire RUN collection, then apply it to 'replaced'
-                    # here.
-                    raise NotImplementedError("Support for --prune-replaced is not yet implemented.")
+                if args.prune_replaced == "unstore":
+                    # Remove datasets from datastore
+                    with butler.transaction():
+                        refs = butler.registry.queryDatasets(..., collections=replaced, deduplicate=True)
```
TallJimbo, line 359: Adding `deduplicate` when there's only one collection is unnecessary - that just invokes the "return the dataset from the first collection in the ordered list" logic, rather than doing a SELECT DISTINCT or something.  It should be harmless, but I should make sure it's truly a no-op (rather than a more complex query for no reason) when there is only one collection; I'm not sure that's the case now.  But I see that `Butler.pruneCollection` does this too, so it's probably my fault for giving you a bad example to learn from.

I wonder if it would be better to modify `Butler.pruneCollection` to accept `unstore=True, purge=False` on a `RUN` so we could use that here; I think that could reduce the amount of total logic this operation goes through.

TallJimbo (approved):
One comment on whether we should move some logic to `daf_butler`; happy to defer that to another ticket.  Thanks for the excellent tests, and I'm sorry I didn't leave you better tests to start from back when I added `--replace-run`.

python/lsst/ctrl/mpexec/cmdLineFwk.py
andy-slac: I removed `deduplicate`, indeed I was pasting without much thinking here.

andy-slac: For `pruneCollection` update I'm also happy to defer to a next ticket.

PR #73, opened 2020-09-14
DM-26728: Implement resource cleanup after job execution

andy-slac:
When child process terminates it needs a special method call on a
Process object. Also added delayed cleanup for processes that are killed
but may not stop immediately.

### DIFF
--- a/tests/test_executors.py
+++ b/tests/test_executors.py
@@ -24,6 +24,7 @@
 
 import logging
 from multiprocessing import Manager
+import psutil
 import time
 from types import SimpleNamespace
 import unittest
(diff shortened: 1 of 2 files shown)


tests/test_executors.py
```diff
@@ -293,6 +294,27 @@ def test_mpexec_failure_failfast(self):
...
+        num_fds_0 = this_proc.num_fds()
+
+        # run in multi-process mode, the order of results is not defined
+        qexec = QuantumExecutorMock(mp=True)
+        mpexec = MPGraphExecutor(numProc=3, timeout=100, quantumExecutor=qexec)
+        mpexec.execute(qgraph, butler=None)
+
+        num_fds_1 = this_proc.num_fds()
+        # they should be the same but allow small growth just in case
+        self.assertLess(num_fds_1 - num_fds_0, 5)
```
timj, line 316: Can you note somewhere what this test gave as the num_fds without your fix? From this line I can't tell where +5 is reasonable or not.

andy-slac: I'll add a comment, without fix it "leaked" 1 file descriptor per sub-process/quantum, totals 20 in this case.

timj (approved):

PR #76, opened 2020-09-16
DM-26645: Add a default value for pipetask2 -j option

n8pease:


andy-slac (approved):
Looks good, sorry I missed clobber_partial_outputs.

PR #78, opened 2020-09-21
DM-26726: Freeze Config prior to running the quantum

timj:


natelust (approved):

python/lsst/ctrl/mpexec/singleQuantumExecutor.py
```diff
@@ -246,8 +249,7 @@ def runQuantum(self, task, quantum, taskDef, butler):
         butlerQC = ButlerQuantumContext(butler, quantum)
 
         # Get the input and output references for the task
-        connectionInstance = task.config.connections.ConnectionsClass(config=task.config)
-        inputRefs, outputRefs = connectionInstance.buildDatasetRefs(quantum)
+        inputRefs, outputRefs = taskDef.connections.buildDatasetRefs(quantum)
```
timj, line 252: @andy-slac I'm not sure if this change is controversial? We have access to the TaskDef so in theory we don't need to recreate the connections object here. I assume the main worry is that by this point the TaskDef and Task can be out of sync somehow?

andy-slac: I am not sure now where the original code comes from, but this change looks OK to me.

natelust: I think I remember that existing prior to TaskDefs carrying around a connections instance, and think it never got cleaned up. This change looks fine.

PR #83, opened 2020-09-28
DM-26739: Add task label to long log message

timj:


andy-slac (approved):
Looks good.

python/lsst/ctrl/mpexec/singleQuantumExecutor.py
```diff
@@ -98,29 +98,27 @@ def execute(self, taskDef, quantum, butler):
...
         """
         # include input dataIds into MDC
         dataIds = set(ref.dataId for ref in chain.from_iterable(quantum.inputs.values()))
         if dataIds:
             if len(dataIds) == 1:
-                Log.MDC("LABEL", str(dataIds.pop()))
+                Log.MDC("LABEL", f"{taskDef.label}: {dataIds.pop()}")
             else:
-                Log.MDC("LABEL", '[' + ', '.join([str(dataId) for dataId in dataIds]) + ']')
+                Log.MDC("LABEL", f'{taskDef.label}:[' + ', '.join([str(dataId) for dataId in dataIds]) + ']')
```
ktlim, line 121: Not a big deal, but why the space after the colon above but not here?  Might make it messier to extract stuff automatically from the logs to have these inconsistent.

timj: It was a mistake. Fixed and removed the space in the first one.

timj: Although extraction of dataIds is going to need a tricky regex, especially if the dataId values include a `]`.

ktlim: This is where JSON logs can help...  Still, the `)(` in the longlog format should be good enough.

timj: Before my SSD died last night and wiped the code out, I was going to see if I can serialize those dataIds into Json form in the log. I'll try it again.

timj:
@andy-slac and @ktlim can you take another look please? I've switched to using JSON to serialize the dataId and removed the special case for single dataId.

andy-slac (approved):
Looks OK.

ktlim:
Well, I was actually suggesting having a mode where the entire log message (including timestamps, log level, etc.) is JSON so that no parsing/regexps are necessary.

If the dataId isn't dict-like enough for JSON to serialize it without help, you might be better off with the original `str()`, especially since you're still tacking on the `{taskdef.label}:`.

timj:
I'm not going to tackle full JSON-ification of logging here. Using the single quantum dataId simplifies the code a lot more so I'll do a final tweak and remove the json.dumps.

PR #87, opened 2020-10-06
DM-26483: Add wildcards to --show history for config name matching

timj:


python/lsst/ctrl/mpexec/cmdLineFwk.py
```diff
@@ -752,27 +752,37 @@ def _showConfigHistory(self, pipeline, showArgs):
...
-                sys.exit(1)
 
-            if hasattr(hconfig, cname):
-                print(f"### Configuration field for task `{taskDef.label}'")
-                print(pexConfig.history.format(hconfig, cname))
-                found = True
+            config = taskDef.config
+
+            # Look for any matches in the config hierarchy for this name
+            for nmatch, thisPattern in enumerate(fnmatch.filter(config.names(), pattern)):
```
ktlim, line 761: Isn't it more like `thisName` than `thisPattern`?

python/lsst/ctrl/mpexec/cmdLineFwk.py
```diff
@@ -752,27 +752,37 @@ def _showConfigHistory(self, pipeline, showArgs):
...
+                        hconfig = getattr(hconfig, cpt)
+                    except AttributeError:
+                        config_path = ".".join(["config"] + cpath[:i])
+                        print(f"Error: Unable to extract {cname} from {config_path} "
+                              f"from task {taskDef.label}", file=sys.stderr)
+                        hconfig = None
+
+                # Sometimes we end up with a non-Config so skip those
+                if isinstance(hconfig, pexConfig.Config) and hasattr(hconfig, cname):
+                    print(f"### Configuration field for task `{taskDef.label}'")
```
ktlim, line 783: I guess this matches the prior output, but it seems weird to have repeated identical section headers like this for a given task.

ktlim (approved):

python/lsst/ctrl/mpexec/cmdLineFwk.py
timj: I think I'm fine with it repeating the task name each time. Otherwise you have to scroll up to find out which task's config you are looking at.

PR #88, opened 2020-10-16
DM-26407: Update unit test for changes in Registry factory method.

andy-slac:


tests/test_cmdLineFwk.py
```diff
@@ -95,7 +95,8 @@ def makeSQLiteRegistry(create=True):
         config = RegistryConfig()
         config["db"] = uri
         if create:
-            Registry.fromConfig(config, create=True)
+            dimensionConfig = DimensionConfig()
+            Registry.createFromConfig(config, dimensionConfig=dimensionConfig)
```
timj, line 99: can't createFromConfig default to doing the right thing?  Can't we have it default to `None` and have that mean "use a default DimensionConfig constructed from daf_butler dimensions.yaml"?  I don't understand why we have to force that code on the caller.

andy-slac: Do you mean making `dimensionConfig` optional? The `fromConfig` method had a `config` as required argument, so I kept `config` as a required argument for `createFromConfig` as well, and for symmetry `dimensionConfig` is required too, I don't see much difference between RegistryConfig and DimensionConfig in this case.

andy-slac: If I make `dimensionConfig` optional should I also make `config` optional too?

timj: I think making the tests create a DimensionConfig knowing that they want to just use the defaults is not helpful and it seems clearer to me to declare up front that you want to use the default dimensions.

I think I could probably go either way on `config` argument. If people are always tweaking the defaults then we don't gain anything by letting it default.

andy-slac: OK, I decided to make them both optional. I prefer symmetry, breaking it spontaneously could create huge mass/inertia in some universe 🙂

TallJimbo (approved):

PR #94, opened 2020-11-04
Fix failfast unit test timing issue (DM-27381)

andy-slac:
Increases time for which a long running task runs so that it does not
finish before failing task fails.


brianv0 (commented):
Looks good!

PR #98, opened 2020-11-17
DM-27500: add a comment in butler & pipetask subcommand about more options in butler --help

n8pease:
PipetaskCommand class allows pipetask-specific overrides of
subcommand behaviors.
This adds a comment to the end of every subcommand's help text,
that more options are available using "pipetask --help"

Removes the --log-level option from subcommands.

There was other help text in the epilog that was specific to
prototype development of chained subcommands, and should have
been removed when we did not use that feature. It is now
removed.

### DIFF
--- a/python/lsst/ctrl/mpexec/cli/utils.py
+++ b/python/lsst/ctrl/mpexec/cli/utils.py
@@ -24,6 +24,7 @@ import collections
 import re
 
 from lsst.daf.butler.cli.opt import (config_file_option, config_option)
+from lsst.daf.butler.cli.utils import MWCommand
 from lsst.obs.base.cli.opt import instrument_option
 from .opt import (delete_option,
                   task_option)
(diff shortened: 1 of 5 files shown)


python/lsst/ctrl/mpexec/cli/utils.py
```diff
@@ -137,3 +138,20 @@ def makePipelineActions(args,
         elif args[i] in instrumentFlags:
             pipelineActions.append(_ACTION_ADD_INSTRUMENT(args[i+1]))
     return pipelineActions
+
+
+class PipetaskCommand(MWCommand):
```
timj, line 143: Given that the only difference between PipetaskCommand and ButlerCommand is the one string, could we set up ButlerCommand with an epilog attribute that says "See butler --help..." and then for PipetaskCommand it's a subclass of ButlerCommand that consists of a single override of the {{epilog}} value? That will stop us duplicating all the code here.

timj (approved):

python/lsst/ctrl/mpexec/cli/utils.py
n8pease: I was worried about getting too cute trying to make them the same, but your suggestion seems like a good compromise, will do.

PR #99, opened 2020-11-17
Use multiprocessing context with fork method

brianv0:
Supports python 3.8 on mac, which defaults to spawn method.


andy-slac (approved):
Looks good.

PR #100, opened 2020-11-18
DM-27653: output.chain[0] now only returns a str not tuple so change assignement

timj:


andy-slac:
Maybe also change type annotation at line 109 to say that `chain` is a list of strings?

timj:
I'm sure this is where @TallJimbo would let us know that type annotations being used would have told us about this problem up front.

TallJimbo (approved):
That thought about type checking did cross my mind :slightly_smiling_face: .  Looks good, and thanks for taking care of it!

PR #101, opened 2020-11-19
DM-27118: Add support for spawn start method

andy-slac:
Pipetask has new option --start-method which tkaes the name of the start
method from multiprocessing (spawn, fork, forkserver). Default method is
now set as "spawn" on MacOS and "fork" on Linux. Some care was needed
when unpickling dimensions, this has to be done after unpickling
registry. Unpickling is ordered manually now. As usual logging causes
issues, with spawn it needs to be re-initialized in every child, added
support for recording and replaying configuration (part of that is in
`daf_butler`).

### DIFF
--- a/python/lsst/ctrl/mpexec/mpGraphExecutor.py
+++ b/python/lsst/ctrl/mpexec/mpGraphExecutor.py
@@ -28,6 +28,7 @@ from enum import Enum
 import logging
 import multiprocessing
 import pickle
+import sys
 import time
 
 from lsst.pipe.base.graph.graph import QuantumGraph
@@ -37,6 +38,7 @@ from lsst.pipe.base.graph.graph import QuantumGraph
 # -----------------------------
 from .quantumGraphExecutor import QuantumGraphExecutor
 from lsst.base import disableImplicitThreading
+from lsst.daf.butler.cli.cliLog import CliLog
 
 _LOG = logging.getLogger(__name__.partition(".")[2])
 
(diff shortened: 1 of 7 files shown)


python/lsst/ctrl/mpexec/mpGraphExecutor.py
```diff
@@ -236,18 +246,29 @@ class MPGraphExecutor(QuantumGraphExecutor):
...
         self.numProc = numProc
         self.timeout = timeout
         self.quantumExecutor = quantumExecutor
         self.failFast = failFast
         self.executionGraphFixup = executionGraphFixup
 
+        # By default use spawn for MacOS and fork for Linux
+        if startMethod is None:
+            methods = dict(linux="fork", darwin="spawn")
+            startMethod = methods.get(sys.platform)
```
timj, line 268: Maybe add a default value for the `get` of "spawn" in case you have an unexpected `sys.platform` value?

timj (approved):

python/lsst/ctrl/mpexec/mpGraphExecutor.py
andy-slac: Unexpected platform will get `None` which will use default `multiprocessing` method. I think that should work better because `spawn` may not work for that unknown platform.

andy-slac: I'll expand the comment above to mention that.

PR #103, opened 2020-11-23
DM-27667: Add pipetask option to execute graph subset

andy-slac:
New option `--qgraph-node-id` is added which takes a bunch of integers
for node numbers to execute from a full graph. Another new option
`--qgraph-id` can be used to specify graph ID to verify it after loading
from pickle file.

### DIFF
--- a/python/lsst/ctrl/mpexec/cli/opt/options.py
+++ b/python/lsst/ctrl/mpexec/cli/opt/options.py
@@ -24,6 +24,38 @@ import click
 from lsst.daf.butler.cli.utils import MWOptionDecorator, MWPath, split_commas, unwrap
 
 
+def _split_commas_int(context, param, values):
+    """Special callback that handles comma-separated list of integers.
+
+    Parameters
+    ----------
+    context : `click.Context` or `None`
+        The current execution context.
+    param : `click.core.Option` or `None`
+        The parameter being handled.
+    values : `list` [ `str` ]
+        All the values passed for this option. Strings may contain commas,
+        which will be treated as delimiters for separate values.
+
+    Returns
+    -------
+    numbers : `tuple` [ `int` ]
+        The passed in values separated by commas and combined into a single
+        list.
+    """
+    def _to_int(value):
+        """Convert string to integer, handle errors."""
+        try:
+            return int(value)
+        except ValueError:
+            raise click.BadParameter(f"'{value}' is not a valid integer", context, param)
+
+    values = split_commas(context, param, values)
+    if values is None:
+        return values
+    return tuple(_to_int(value) for value in values)
+
+
 butler_config_option = MWOptionDecorator("-b", "--butler-config",
                                          help="Location of the gen3 butler/registry config file.")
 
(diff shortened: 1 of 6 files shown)


python/lsst/ctrl/mpexec/cli/opt/options.py
```diff
@@ -22,6 +22,38 @@
 import click
 
 from lsst.daf.butler.cli.utils import MWOptionDecorator, MWPath, split_commas, unwrap
+from lsst.daf.butler.core.utils import iterable
+
+
+def _split_commas_int(context, param, values):
```
timj, line 28: Why has this been copied from daf_butler? Don't we have the ability to use the daf_butler callback?

cc/ @n8pease

andy-slac: It's not an exact copy, daf_butler's one splits things into strings, this one make integers instead. But I should have probably called `split_commas` from this method instead of re-implementing its logic.

andy-slac: Re-reading the question again - I guess we could update `split_commas` to do type conversion, but my general problem is that `click` callbacks do not seem to be a good match for that problem. AFAIK callbacks happen after type conversion is already applied to argument, and in our case of comma-separated lists it cannot even work. What I mean is that if you declare an option to have type `int` than you cannot pass comma-separated string to it because it will try to convert that string to `int` and will fail. My current solution is not to declare type of an option (essentially making it `str`) but cheat in the callback and return integers from that. I don't know if this is even allowed, docs don't say if this is acceptable, so there is no guarantee that it will not break.

python/lsst/ctrl/mpexec/cli/script/qgraph.py
```diff
@@ -42,6 +42,12 @@ def qgraph(pipelineObj, qgraph, skip_existing, save_qgraph, save_single_quanta,
     qgraph : `str` or `None`
         URI location for a serialized quantum graph definition as a pickle
         file. If this option is not None then `pipeline` should be `None`.
+    qgraph_id : `str` or `None`
+        Quantum graph identifier, graph loaded from a pickle file is required
+        to have this identifier. Ignored if graph is not loaded from a file.
```
MichelleGower, line 47: Did you want to change the wording here to match the updated help wording?

MichelleGower (approved):
Tested the changes and they worked as expected.  Only 1 very minor question about a docstring.  Ready to merge.

python/lsst/ctrl/mpexec/cli/opt/options.py
andy-slac: Just for record - I reimplemented my `_split_commas_int` on top of `split_commas`, for better unification of the latter we may need some thinking, so I want to keep `_split_commas_int` here for now.

python/lsst/ctrl/mpexec/cli/script/qgraph.py
andy-slac: OK, will do.

PR #108, opened 2020-12-07
DM-27835: pipetask run should always print stack trace and exception details

n8pease:


tests/test_cmdLineFwk.py
```diff
@@ -153,16 +157,22 @@ def _makeArgs(registryConfig=None, **kwargs):
...
+        """Fake 'pipetask run' command for gathering input arguments."""
+        mock(**kwargs)
+
     runner = click.testing.CliRunner()
-    result = runner.invoke(pipetaskCli, ["run", "--call-mocker"])
+    result = runner.invoke(run)
     if result.exit_code != 0:
-        raise RuntimeError(f"Failure getting default args from 'pipetask run': {result}")
-    _, args = Mocker.mock.call_args
+        raise RuntimeError("Failure getting default args from 'pipetask run'.")
```
andy-slac, line 173: Did you drop "{result}" from message intentionally? It was helpful for me to debug things, without that I could not figure out what was happening in the code.

andy-slac (approved):
Looks good.

tests/test_cmdLineFwk.py
n8pease: I'll clean this up some to be more clear; the message is misleading; it's not calling "pipetask run", it's calling the "fake" run function (defined a few lines before this), which has the same options as the "real" run function. 

It's worth keeping `result` in too, in case the options fail it may be well explained by result. Or, you may just have to pdb (sorry 🙃), but hopefully it all Just Works all the time!

PR #112, opened 2021-01-20
DM-28418: Improve scheduler performance for large graph

andy-slac:
Multiprocessing scheduler did not scale very well with the number of
nodes in a quantum graph. Re-designed data structures to avoid looping
through the list of nodes on every iteration.

### DIFF
--- a/python/lsst/ctrl/mpexec/mpGraphExecutor.py
+++ b/python/lsst/ctrl/mpexec/mpGraphExecutor.py
@@ -64,8 +64,13 @@ class _Job:
     def __init__(self, qnode):
         self.qnode = qnode
         self.process = None
-        self.state = JobState.PENDING
+        self._state = JobState.PENDING
         self.started = None
+
+    @property
+    def state(self):
+        """Job processing state (JobState)"""
+        return self._state
 
     def start(self, butler, quantumExecutor, startMethod=None):
         """Start process which runs the task.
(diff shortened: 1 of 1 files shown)


python/lsst/ctrl/mpexec/mpGraphExecutor.py
```diff
@@ -157,57 +162,72 @@ class _JobList:
...
+        job._state = state
+        if state == JobState.FINISHED:
+            self.finishedNodes.add(qnode)
+        elif state == JobState.FAILED:
+            self.failedNodes.add(qnode)
+        elif state == JobState.FAILED_DEP:
+            self.failedNodes.add(qnode)
+        elif state == JobState.TIMED_OUT:
+            self.failedNodes.add(qnode)
+            self.timedOutNodes.add(qnode)
```
timj, line 230: can you put an else at the end of this that raises? I know that it shouldn't get to it if the assert works above but in the future someone might clean something up that removes it.

python/lsst/ctrl/mpexec/mpGraphExecutor.py
```diff
@@ -402,46 +422,54 @@ def _executeQuantaMP(self, graph, butler):
...
-                    if len(jobs.running()) < self.numProc:
-                        _LOG.debug("Sumbitting %s", job)
-                        job.start(butler, self.quantumExecutor, self.startMethod)
+            if len(jobs.running) < self.numProc:
+                for job in jobs.pending:
+                    jobInputNodes = graph.determineInputsToQuantumNode(job.qnode)
+                    if jobInputNodes <= jobs.finishedNodes:
+                        # all dependencies have completed, can start new job
+                        if len(jobs.running) < self.numProc:
+                            _LOG.debug("Sumbitting %s", job)
```
timj, line 441: Sumbitting?

timj (approved):

PR #115, opened 2021-02-12
DM-28598: Fix pipetask command-line utility doc generation.

TallJimbo:
The rST content was severely bitrotted.  I'm not sure what was going on with the conf.py, but sphinx couldn't find its click extensions
without this change, which is just mindless cargo-culting from daf_butler.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -3,11 +3,10 @@
 This configuration only affects single-package Sphinx documentation builds.
 """
 
-from documenteer.sphinxconfig.stackconf import build_package_configs
-import lsst.ctrl.mpexec
+from documenteer.conf.pipelinespkg import *  # noqa: F403, import *
 
-
-_g = globals()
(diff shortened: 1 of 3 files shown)


timj (approved):
The conf.py was the old one for the old documenteer so the change to that is good.

jonathansick (changes requested):
Yep, these changes are fine.

One bit of feedback is that the practice of putting the command-line reference directly on the module's homepage isn't consistent with our documentation design. Instead, you want each command-line reference on its own page, and link to those command-line references using a toctree in the `index.rst` file using the "Command line scripts" heading.

TallJimbo:
Merging before the `pipelines.lsst.io` component of the ticket is complete so I can more easily test that they work together, by getting this into a daily.  (@jonathansick 's requested change has been addressed, and he signed off on Jira.)

PR #117, opened 2021-03-12
DM-29203: Recursively flatten output CHAINED collections.

TallJimbo:
This does a better (albeit still incomplete) job of recording what was actually used in a processing run, as opposed to the latest version of something like what was actually used in the run.

### DIFF
--- a/python/lsst/ctrl/mpexec/cmdLineFwk.py
+++ b/python/lsst/ctrl/mpexec/cmdLineFwk.py
@@ -199,7 +199,11 @@ class _ButlerFactory:
             self.outputRun = None
         else:
             raise ValueError("Cannot write without at least one of (--output, --output-run).")
-        self.inputs = tuple(CollectionSearch.fromExpression(args.input)) if args.input else ()
+        # Recursively flatten any input CHAINED collections.  We do this up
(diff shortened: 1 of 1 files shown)


python/lsst/ctrl/mpexec/cmdLineFwk.py
```diff
@@ -199,7 +199,13 @@ def __init__(self, registry: Registry, args: argparse.Namespace, writeable: bool
             self.outputRun = None
         else:
             raise ValueError("Cannot write without at least one of (--output, --output-run).")
-        self.inputs = tuple(CollectionSearch.fromExpression(args.input)) if args.input else ()
+        # Recursively flatten any input CHAINED collections.  We do this up
+        # front so we can tell if the user passes the same inputs on subsequent
+        # calls, even though we also flatten when we define the output CHAINED
+        # collection.
+        self.inputs = CollectionSearch.fromExpression(
+            list(registry.queryCollections(args.input, flattenChains=True)) if args.input else ()
```
andy-slac, line 207: `fromExpression` accepts iterables so `list()` should not be needed here, or if you want to make a list then `else` should also produce a list, not tuple?

andy-slac (approved):
Looks good, one minor comment.

PR #123, opened 2021-05-11
DM-30148: PipelineTasks use wrong label as name

kfindeisen:
This PR fixes a name collision inadvertently introduced in #122, and adds unit tests to catch similar bugs in the future.

### DIFF
--- a/tests/test_taskFactory.py
+++ b/tests/test_taskFactory.py
@@ -1,0 +1,166 @@
+# This file is part of ctrl_mpexec.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+import shutil
+import tempfile
+import unittest
+
+import lsst.afw.table as afwTable
+import lsst.daf.butler.tests as butlerTests
+import lsst.pex.config as pexConfig
+from lsst.pipe.base import PipelineTaskConfig, PipelineTaskConnections
+from lsst.pipe.base.configOverrides import ConfigOverrides
+from lsst.pipe.base import connectionTypes
+
+from lsst.ctrl.mpexec import TaskFactory
+
+
+class FakeConnections(PipelineTaskConnections, dimensions=set()):
+    initInput = connectionTypes.InitInput(name="fakeInitInput", doc="", storageClass="SourceCatalog")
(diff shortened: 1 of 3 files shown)


tests/test_taskFactory.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+def mockTaskClass():
+    """A class placeholder that records calls to __call__.
+    """
+    mock = unittest.mock.Mock(__name__="_TaskMock", _DefaultName="FakeTask", ConfigClass=FakeConfig)
+    return mock
+
+
+class TaskFactoryTestCase(unittest.TestCase):
```
parejkoj, line 54: Please add a docstring: it's not obvious to me what a "TaskFactory" test would be testing. I think it's testing `TaskFactory.makeTask()` with a variety of combinations of optional arguments?

tests/test_taskFactory.py
```diff
@@ -0,0 +1,162 @@
...
+        self.factory.makeTask(taskClass=self.tracker, label=None, config=None,
+                              overrides=None, butler=None)
+        self.tracker.assert_called_with(config=FakeConfig(), initInputs=None, name=None)
+
+    def testAllOptions(self):
+        butler = self._tempButler()
+        self.factory.makeTask(taskClass=self.tracker, label="no-name", config=self._alteredConfig(),
+                              overrides=self._overrides(), butler=butler)
+        catalog = butler.get("fakeInitInput")  # Copies of _dummyCatalog are identical but not equal
+        # Overrides ignored
```
parejkoj, line 110: Please expand on this comment: I think you mean "when `config` is passed, `overrides` is ignored", but I'm not sure.

tests/test_taskFactory.py
```diff
@@ -0,0 +1,162 @@
...
+        butler.put(catalog, "fakeInput")
+        butler.put(catalog, "fakeOutput")
+        return butler
+
+    def testNoOptions(self):
+        self.factory.makeTask(taskClass=self.tracker, label=None, config=None,
+                              overrides=None, butler=None)
+        self.tracker.assert_called_with(config=FakeConfig(), initInputs=None, name=None)
+
+    def testAllOptions(self):
```
parejkoj, line 105: This is a test of `makeTask` called with all kwargs? Maybe rename it or add a docstring.

tests/test_taskFactory.py
```diff
@@ -0,0 +1,162 @@
...
+        catalog = butler.get("fakeInitInput")  # Copies of _dummyCatalog are identical but not equal
+        self.tracker.assert_called_with(config=self._alteredConfig(),
+                                        initInputs={'initInput': catalog},
+                                        name=None)
+
+    def testOverridesButler(self):
+        butler = self._tempButler()
+        self.factory.makeTask(taskClass=self.tracker, label=None, config=None,
+                              overrides=self._overrides(), butler=butler)
+        config = self._alteredConfig()
```
parejkoj, line 157: I think this should be `config = FakeConfig()`, since we don't pass the alteredConfig at all. I think it doesn't matter in practice here, but if someone were to add another config option, it might?

And the same in `testNameOverrides`, now that I notice it.

parejkoj (approved):
A PR that is almost all tests to find and fix a small bit of code and add coverage to un-covered code. Wholehearted approval!

A few docs-related suggestions, but otherwise good.

tests/test_taskFactory.py
parejkoj: Hmmm, maybe the fact that `TaskFactory` itself is short on documentation is really my problem? I was looking at the test without knowing anything about TaskFactory: I didn't know it only had one method.

tests/test_taskFactory.py
kfindeisen: `testOnlyMandatoryArg` and `testAllArgs`?

Note that "kwargs" is kind of a pointless term here, since none of the parameters have defaults.

PR #125, opened 2021-06-08
DM-30653: Add v22 changes and enable towncrier

timj:


TallJimbo (approved):

doc/changes/README.rst
```diff
@@ -0,0 +1,24 @@
+Recording Changes
+=================
+
+This directory contains "news fragments" which are small files containing text that will be integrated into release notes.
+The files can be in restructured text format or plain text.
```
andy-slac, line 5: Does it support Markdown? Some people are more comfortable with Markdown than restructured text.

timj: Yes, you can just do `DM-1234.feature.md` and it will do the right thing (I had an example of that in the daf_butler fragments).

PR #126, opened 2021-06-09
DM-30649: required/optional inputs, better skip-existing and clobber-outputs, and NoWorkFound propagation

TallJimbo:


python/lsst/ctrl/mpexec/singleQuantumExecutor.py
```diff
@@ -144,8 +180,10 @@ def checkExistingOutputs(self, quantum, butler, taskDef):
         Returns
         -------
         exist : `bool`
-            True if all quantum's outputs exist in a collection and
-            ``skipExisting`` is True, False otherwise.
+            True if a previous execution of this quanta appears to have
+            completed successfully (either because metadata was written or all
+            datasets were written) and ``skipExisting`` is `True`, `False`
```
natelust, line 185: im not sure the bit about skipExisting is clear in this doc string

python/lsst/ctrl/mpexec/singleQuantumExecutor.py
```diff
@@ -155,6 +193,13 @@ def checkExistingOutputs(self, quantum, butler, taskDef):
         collection = butler.run
         registry = butler.registry
 
+        if self.skipExisting and taskDef.metadataDatasetName is not None:
+            # Metadata output exists; this is sufficient to assume the previous
+            # run was successful and should be skipped.
+            if (ref := butler.registry.findDataset(taskDef.metadataDatasetName, quantum.dataId)) is not None:
```
natelust, line 199: Why are we looking this up? Shouldn't this be known prior to execution time? Not if it exists, because you might be re-running the graph, but the reference should be known right?

python/lsst/ctrl/mpexec/singleQuantumExecutor.py
```diff
@@ -27,15 +27,23 @@
...
 from itertools import chain
+import sys
 import time
 
 # -----------------------------
 #  Imports for other modules --
 # -----------------------------
 from .quantumGraphExecutor import QuantumExecutor
 from lsst.log import Log
+from lsst.daf.base import PropertyList, PropertySet
```
timj, line 38: Why are we now adding a dependency on daf_base when we are trying to get gen3 middleware to stop using it?

python/lsst/ctrl/mpexec/singleQuantumExecutor.py
```diff
@@ -271,20 +344,38 @@ def runQuantum(self, task, quantum, taskDef, butler):
...
+            _LOG.info("Task '%s' on quantum %s exited early: %s",
+                      taskDef.label, quantum.dataId, str(err))
+            pass
+        except RepeatableQuantumError as err:
+            if self.exitOnKnownError:
+                _LOG.warning("Caught repeatable quantum error for %s (%s):", taskDef, quantum.dataId)
+                _LOG.warning(err, exc_info=True)
+                sys.exit(err.EXIT_CODE)
+            else:
+                raise
```
natelust, line 362: Im not sure what the difference is here, this will bring it down in their case right? Is it that EXIT_CODE is something other than say 1?

python/lsst/ctrl/mpexec/singleQuantumExecutor.py
```diff
@@ -271,20 +344,38 @@ def runQuantum(self, task, quantum, taskDef, butler):
...
+        except InvalidQuantumError as err:
+            _LOG.fatal("Invalid quantum error for %s (%s): %s", taskDef, quantum.dataId)
+            _LOG.fatal(err, exc_info=True)
+            sys.exit(err.EXIT_CODE)
```
natelust, line 366: Should this be the same exit code here?

python/lsst/ctrl/mpexec/singleQuantumExecutor.py
TallJimbo: Because tasks do still use it, and it's not hard to back out later.

python/lsst/ctrl/mpexec/singleQuantumExecutor.py
```diff
@@ -55,24 +63,35 @@ class SingleQuantumExecutor(QuantumExecutor):
...
-        self.clobberPartialOutputs = clobberPartialOutputs
+        self.clobberOutputs = clobberOutputs
+        self.exitOnKnownError = exitOnKnownError
 
     def execute(self, taskDef, quantum, butler):
 
         startTime = time.time()
 
+        # Save detailed resource usage before task start to metadata.
+        quantumMetadata = PropertyList()
```
timj, line 92: This does not need to be a property list does it?

python/lsst/ctrl/mpexec/singleQuantumExecutor.py
```diff
@@ -83,7 +102,20 @@ def execute(self, taskDef, quantum, butler):
...
+        except NoWorkFound as exc:
+            _LOG.info("Nothing to do for task '%s' on quantum %s; skipping: %s",
+                      taskDef.label, quantum.dataId, str(exc))
+            # Make empty metadata that looks something like what a do-nothing
+            # task would write (but we don't bother with empty nested
+            # PropertySets for subtasks).  This is slightly duplicative with
+            # logic in pipe_base that we can't easily call from here; we'll fix
+            # this on DM-29761.
+            fullMetadata = PropertySet()
+            fullMetadata.set(taskDef.label, PropertyList())
```
timj, line 116: Please don't use `set` with a `PropertyX` any more. This should all work with:

```
fullMetadata[taskeDef.label] = {}
fullMetadata["quantum"] = quantumMetadata
```
and the `__setitem__` code automatically fixes things up internally.  That way when PropertySet is removed in the line above nothing else needs to be edited.

python/lsst/ctrl/mpexec/singleQuantumExecutor.py
TallJimbo: I'm just doing exactly what `Task` itself does to avoid surprises.

python/lsst/ctrl/mpexec/singleQuantumExecutor.py
TallJimbo: It's just easier to get the data ID and dataset type name separately than to find the right (unresolved) `DatasetRef` in the quantum and pass its data ID and dataset type to `findDataset`.  The behavior is identical the same either way because they have the same information content.

python/lsst/ctrl/mpexec/singleQuantumExecutor.py
TallJimbo: Will reword.

python/lsst/ctrl/mpexec/singleQuantumExecutor.py
TallJimbo: Yes; I'm trying to give BPS a chance to do special things for special exceptions, and that means translating exceptions to exit codes (it's also a convenient way to tell the multiprocessing executor what exception type brought down a worker node, though I'm sure there are other ways to do that).

python/lsst/ctrl/mpexec/singleQuantumExecutor.py
TallJimbo: I don't understand.  Same exit code as what?

natelust: Never mind, I clarified my question on the pipe_base review

python/lsst/ctrl/mpexec/singleQuantumExecutor.py
yalsayyad: Looking at https://lsst.ncsa.illinois.edu/~yusra/download/log.log maybe this needs to reraise its own `NoWorkFound`?

TallJimbo: Yeah, it needs to do something there, but I think `return` early is what we want.

natelust (approved):

PR #127, opened 2021-06-15
Tickets/DM-30702 Add metadata to QuantumGraph

natelust:
## Checklist

- [ x ] ran Jenkins - https://ci.lsst.codes/job/stack-os-matrix/34456/display/redirect
- [ x ] added a release note for user-visible changes to `doc/changes`

### DIFF
--- a/python/lsst/ctrl/mpexec/cmdLineFwk.py
+++ b/python/lsst/ctrl/mpexec/cmdLineFwk.py
@@ -29,7 +29,9 @@ __all__ = ['CmdLineFwk']
 # -------------------------------
 import argparse
 import copy
+import datetime
 import fnmatch
+import getpass
 import logging
 import re
 import sys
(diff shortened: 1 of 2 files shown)


python/lsst/ctrl/mpexec/cmdLineFwk.py
```diff
@@ -539,7 +539,11 @@ def makeGraph(self, pipeline, args):
             # make execution plan (a.k.a. DAG) for pipeline
             graphBuilder = GraphBuilder(registry,
                                         skipExisting=args.skip_existing)
-            qgraph = graphBuilder.makeGraph(pipeline, collections, run, args.data_query)
+            # accumulate metadata
+            metadata = {"input": args.input, "output": args.output, "butler_config": args.butler_config,
+                        "output_run": args.output_run, "extend_run": args.extend_run,
+                        "skip_existing": args.skip_existing}
```
andy-slac, line 545: On JIRA @timj also asked about few other items: time of creation, user name, data query (should be in `args.data_query`). I am somewhat concerned that `butler_config` may point to some temporary location which could disappear, should we be storing registry database URL just in case?

andy-slac (approved):
Looks good, but check my comments, I think we may want few extra things in metadata.

python/lsst/ctrl/mpexec/cmdLineFwk.py
timj: I think the butler URI for the repository itself is probably better than the `ButlerConfig` (or at least we should definitely include the butler URI regardless even if we think we also need the config itself).

andy-slac: `butler_config` here is just a command line argument which is a path/URL, not the whole `ButlerConfig`.

timj: Okay. That's not a good name for it then.

andy-slac: Indeed, historically it comes from `--butler-config` command line option which by default was translated to `butler_config` attribute name by `argparse` (which we don't use any more).

natelust: I changed the key in the metadata to say "butler_argument" to better reflect that it was what was passed on the command line, and side step the name `butler_config`

PR #129, opened 2021-07-16
DM-31120: Set MDC information in butler rather than explicitly lsst.log

timj:
There is also a docstring cleanup.

Depends on lsst/daf_butler#549

## Checklist

- [x] ran Jenkins


andy-slac (approved):
Looks good.

PR #130, opened 2021-07-20
DM-30977: Accumulate log messages per quantum and store in butler

timj:
Relies on lsst/daf_butler#546 and lsst/pipe_base#193

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/ctrl/mpexec/singleQuantumExecutor.py
+++ b/python/lsst/ctrl/mpexec/singleQuantumExecutor.py
@@ -24,11 +24,13 @@ __all__ = ['SingleQuantumExecutor']
 # -------------------------------
 #  Imports of standard modules --
 # -------------------------------
-from collections import defaultdict
 import logging
-from itertools import chain
 import sys
+import tempfile
 import time
+from collections import defaultdict
+from itertools import chain
+from logging import FileHandler
 from typing import List
 
 # -----------------------------
@@ -45,8 +47,16 @@ from lsst.pipe.base import (
     RepeatableQuantumError,
     logInfo,
 )
-from lsst.daf.butler import Quantum, ButlerMDC, NamedKeyDict, DatasetRef, DatasetType, ButlerLogRecordHandler
-
+from lsst.daf.butler import (
+    ButlerLogRecordHandler,
+    ButlerMDC,
+    DatasetRef,
+    DatasetType,
+    FileDataset,
+    JsonFormatter,
+    NamedKeyDict,
+    Quantum,
+)
 # ----------------------------------
 #  Local non-exported definitions --
 # ----------------------------------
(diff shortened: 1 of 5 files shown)


python/lsst/ctrl/mpexec/singleQuantumExecutor.py
```diff
@@ -163,6 +182,24 @@ def setupLogging(self, taskDef, quantum):
 
         ButlerMDC.MDC("LABEL", label)
 
+        # Add the handler to the root logger.
+        # How does it get removed reliably?
```
andy-slac, line 186: Do you mean if writeLogRecords is not called?

python/lsst/ctrl/mpexec/singleQuantumExecutor.py
```diff
@@ -374,6 +411,50 @@ def writeMetadata(self, quantum, metadata, taskDef, butler):
...
+                    f" and execution") from exc
+
+            if isinstance(self.log_handler, ButlerLogRecordHandler):
+                butler.put(self.log_handler.records, ref[0])
+
+                # Clear the records in case the handler is reused.
+                self.log_handler.records.clear()
+            else:
+                if filename is None:
+                    raise RuntimeError("Something bad went wrong extracting filename from"
```
andy-slac, line 443: `Something bad went wrong` - does it mean it's not bad anymore? 😄

andy-slac (approved):
Looks good

python/lsst/ctrl/mpexec/singleQuantumExecutor.py
timj: Yes, maybe I should wrap the entire `execute` method in a `try/finally`.  On the other hand if I do fail to remove the handler eventually you'll get a `ResourceWarning` -- the tests did actually fail until I fixed some of the edge cases for calling writeLogRecords

python/lsst/ctrl/mpexec/singleQuantumExecutor.py
timj: Maybe it should be an assert here because this check is never meant to trigger -- if it does something unexpected happened.

python/lsst/ctrl/mpexec/singleQuantumExecutor.py
andy-slac: Or maybe make `setupLogging` to return a context manager and wrap all code in `with`?

timj: Hmm. Very good point.

timj: I also took the opportunity to add a context manager for the ButlerMDC setting so that it gets reset on exit.  Feel free to take a second look at this PR -- it's going to be a few hours before Jenkins completes.

python/lsst/ctrl/mpexec/singleQuantumExecutor.py
```diff
@@ -77,92 +90,141 @@ class SingleQuantumExecutor(QuantumExecutor):
...
+
+        Notes
+        -----
+        Expected to be used as a context manager to ensure that logging
+        records are inserted into the butler once the quantum has been
+        executed:
+
+        .. code-block:: py
+
+           with self.captureLogging(taskDef, quantum):
```
andy-slac, line 192: Needs a `butler` argument too

andy-slac (approved):
Looks good, one docstring comment.

PR #132, opened 2021-07-29
DM-31255: Fix log-to-butler failures with --skip-existing

andy-slac:
Add special logic to detect cases when we don't want to save log records
to butler. Removes temporary file in that case or when ingest does not
work. When ingest is not supported we now store an empty records object
to make sure that dataset exists for consistency with other cases.

## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/ctrl/mpexec/singleQuantumExecutor.py
+++ b/python/lsst/ctrl/mpexec/singleQuantumExecutor.py
@@ -25,6 +25,7 @@ __all__ = ['SingleQuantumExecutor']
 #  Imports of standard modules --
 # -------------------------------
 import logging
+import os
 import sys
 import tempfile
 import time
@@ -57,6 +58,7 @@ from lsst.daf.butler import (
 )
 from lsst.daf.butler.core.logging import (
     ButlerLogRecordHandler,
+    ButlerLogRecords,
     ButlerMDC,
     JsonLogFormatter,
 )
@@ -65,6 +67,12 @@ from lsst.daf.butler.core.logging import (
 # ----------------------------------
 
 _LOG = logging.getLogger(__name__.partition(".")[2])
+
+
+class _LogCaptureFlag:
+    """Simple flag to enable/disable log-to-butler saving.
+    """
+    store: bool = True
 
 
 class SingleQuantumExecutor(QuantumExecutor):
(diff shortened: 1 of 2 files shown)


python/lsst/ctrl/mpexec/singleQuantumExecutor.py
```diff
@@ -474,12 +498,18 @@ def writeLogRecords(self, quantum, taskDef, butler):
...
+                    # datastore when testing), we store empty list for those
+                    # just to have a dataset. Alternative is to read the file
+                    # as a ButlerLogRecords object and put it.
                     _LOG.info("Log records could not be stored in this butler because the"
-                              " datastore can not ingest files.")
-                    pass
+                              " datastore can not ingest files, empty record list is stored instead.")
+                    records = ButlerLogRecords.from_records([])
+                    butler.put(records, ref[0])
+                    # And remove the file, but ignore errors
```
timj, line 508: Could maybe put this removal in a `finally` clause.

timj (approved):
Looks good to me. Thanks for fixing up my problem.

python/lsst/ctrl/mpexec/singleQuantumExecutor.py
```diff
@@ -450,6 +465,15 @@ def writeLogRecords(self, quantum, taskDef, butler):
             # Remove the handler so we stop accumulating log messages.
             logging.getLogger().removeHandler(self.log_handler)
 
+        if not store:
+            # do not store anything but remove file
+            if filename is not None:
```
timj, line 470: This duplication logic really makes me wish that python had a `if`/`else`/`finally` clause so we could do:
```
if store:
    # store it
else:
    pass
finally:
    # remove file
```

timj: although clearly I'm being silly:

```python
if store and other tests:
    # store it

if filename:
    # remove it
```
would work wouldn't it?

andy-slac: Should work, I'll change that.

andy-slac: There is also case when `ingest` fails with something else than `NotImplementedError`, I think we also need to cleanup in that case.

andy-slac: New commit should fix all problems 🙂

PR #135, opened 2021-08-24
DM-31507 Explicitly run garbage collection

natelust:
Some objects are not cleaned up until garbage collection is run,
which can happen at unpredictable times. Run collection explicitly
to make sure memory is freed. This HOPEFULLY will only be third
party objects, and none of our own.

## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/ctrl/mpexec/mpGraphExecutor.py
+++ b/python/lsst/ctrl/mpexec/mpGraphExecutor.py
@@ -25,6 +25,7 @@ __all__ = ["MPGraphExecutor", "MPGraphExecutorError", "MPTimeoutError"]
 #  Imports of standard modules --
 # -------------------------------
 from enum import Enum
+import gc
 import logging
 import multiprocessing
 import pickle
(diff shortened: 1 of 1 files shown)


python/lsst/ctrl/mpexec/mpGraphExecutor.py
```diff
@@ -348,7 +349,13 @@ def _executeQuantaInProcess(self, graph, butler):
...
         for qnode in graph:
             _LOG.debug("Executing %s", qnode)
-            self.quantumExecutor.execute(qnode.taskDef, qnode.quantum, butler)
+            try:
+                self.quantumExecutor.execute(qnode.taskDef, qnode.quantum, butler)
+            finally:
+                # sqlalchemy has some objects that can last until a garbage
+                # collection cycle is run, which can happen at unpredictable
+                # times, run a collection loop here explicitly.
+                gc.collect()
```
andy-slac, line 358: Running GC frequently can be very expensive, I wonder if we could avoid pessimization by running it conditionally. In my old project when I did something similar I only ran it when virtual mem size increased significantly since last explicit CG. This of course needs some guessing as to what makes a significant increase, in that project threshold was something like 10MB, but it clearly depends on growth pattern.

andy-slac (approved):
Looks good, one suggestion.

PR #136, opened 2021-08-30
DM-31579: Force temp file for log collection to have normal file protections

timj:
tempfile forces restrictive permission on all files and this
can cause read permissions later in the datastore.

## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/ctrl/mpexec/singleQuantumExecutor.py
+++ b/python/lsst/ctrl/mpexec/singleQuantumExecutor.py
@@ -223,6 +223,21 @@ class SingleQuantumExecutor(QuantumExecutor):
                                                   delete=False)
                 self.log_handler = FileHandler(tmp.name)
                 tmp.close()
+
+                # Want to ensure that this temp file has same permissions
+                # as a standard file since at the end this file will be
+                # ingested to datastore. Do this by reading umask
+                # (which requires we set umask and then set it again).
+                # Changing umask is process wide so pick something that
+                # is least dangerous. Other alternatives are to fork a process
+                # that prints the umask. Using transfer=copy does not help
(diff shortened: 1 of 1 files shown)


python/lsst/ctrl/mpexec/singleQuantumExecutor.py
```diff
@@ -223,6 +223,21 @@ def captureLogging(self, taskDef, quantum, butler):
...
+                # as a standard file since at the end this file will be
+                # ingested to datastore. Do this by reading umask
+                # (which requires we set umask and then set it again).
+                # Changing umask is process wide so pick something that
+                # is least dangerous. Other alternatives are to fork a process
+                # that prints the umask. Using transfer=copy does not help
+                # because shutil.copy always copies file permissions.
+                # Alternatively, create a temporary directory and then open
+                # the file inside that directory without using tempfile.
+                umask = os.umask(0o666)
```
ktlim, line 237: This potential race is indeed undesirable.  If you're going to have it, isn't 0o3777 even more safe than 0o666?

You could have a set of fallbacks instead:

1. See if `/proc/self/status` exists and read the `Umask: ` line from that
2. If `/usr/bin/umask` exists, `umask = int(subprocess.check_output("/usr/bin/umask"), base=8)`
3. If this works, use it: `umask = int(subprocess.check_output("umask", shell=True), base=8)`
4. Fall back to your own process.

But that seems like a lot of trouble; after listing those out, the temporary directory solution seems the most portable and the least hassle.

timj: I've rewritten it to use a temp dir.

ktlim (approved):

PR #138, opened 2021-09-13
DM-25013: Improve error messages (ctrl_mpexec)

andy-slac:
Few small improvements to generate better exception messages.

## Checklist

- [X] ran Jenkins

### DIFF
--- a/python/lsst/ctrl/mpexec/cli/cmd/commands.py
+++ b/python/lsst/ctrl/mpexec/cli/cmd/commands.py
@@ -25,6 +25,7 @@ from lsst.daf.butler.cli.opt import (config_file_option,
                                      config_option,
                                      options_file_option)
 from lsst.daf.butler.cli.utils import (MWCtxObj,
+                                       catch_and_exit,
                                        option_section,
                                        unwrap)
 import lsst.obs.base.cli.opt as obsBaseOpts
@@ -70,6 +71,7 @@ def _doBuild(ctx, **kwargs):
 @ctrlMpExecOpts.pipeline_build_options()
 @option_section(sectionText="")
 @options_file_option()
+@catch_and_exit
 def build(ctx, **kwargs):
     """Build and optionally save pipeline definition.
 
@@ -86,6 +88,7 @@ def build(ctx, **kwargs):
 @ctrlMpExecOpts.butler_options()
 @option_section(sectionText="")
 @options_file_option()
+@catch_and_exit
 def qgraph(ctx, **kwargs):
     """Build and optionally save quantum graph.
     """
@@ -95,6 +98,7 @@ def qgraph(ctx, **kwargs):
 
 @click.command(cls=PipetaskCommand, epilog=epilog)
 @ctrlMpExecOpts.run_options()
+@catch_and_exit
 def run(ctx, **kwargs):
     """Build and execute pipeline and quantum graph.
     """
--- a/python/lsst/ctrl/mpexec/cli/opt/optionGroups.py
+++ b/python/lsst/ctrl/mpexec/cli/opt/optionGroups.py
@@ -84,7 +84,7 @@ class butler_options(OptionGroup):  # noqa: N801
     def __init__(self):
         self.decorators = [
             option_section(sectionText="Data repository and selection options:"),
-            ctrlMpExecOpts.butler_config_option(),
+            ctrlMpExecOpts.butler_config_option(required=True),
             ctrlMpExecOpts.input_option(),
             ctrlMpExecOpts.output_option(),
             ctrlMpExecOpts.output_run_option(),
--- a/python/lsst/ctrl/mpexec/cmdLineFwk.py
+++ b/python/lsst/ctrl/mpexec/cmdLineFwk.py
@@ -194,6 +194,8 @@ class _ButlerFactory:
             self.outputRun = _OutputRunCollectionInfo(registry, args.output_run)
         elif self.output is not None:
             if args.extend_run:
+                if not self.output.chain:
+                    raise ValueError("Cannot use --extend-run option with non-existing or empty output chain")
                 runName = self.output.chain[0]
             else:
                 runName = "{}/{}".format(self.output, Instrument.makeCollectionTimestamp())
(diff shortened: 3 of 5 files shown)


python/lsst/ctrl/mpexec/cli/cmd/commands.py
```diff
@@ -75,7 +76,11 @@ def build(ctx, **kwargs):
 
     This does not require input data to be specified.
     """
-    _doBuild(ctx, **kwargs)
+    try:
+        _doBuild(ctx, **kwargs)
+    except Exception:
+        traceback.print_exc()
+        ctx.exit(1)
```
timj, line 83: Interesting. I did a bit of an experiment before and after because I wasn't initially clear what was happening:

```
$ butler create REPO
Traceback (most recent call last):
  File "/Users/timj/work/lsstsw3/build/daf_butler_test/bin/butler", line 28, in <module>
    sys.exit(main())
  File "/Users/timj/work/lsstsw3/build/daf_butler_test/python/lsst/daf/butler/cli/butler.py", line 334, in main
    return cli()
  File "/Users/timj/work/lsstsw3/miniconda/envs/lsst-scipipe-0.6.0/lib/python3.8/site-packages/click/core.py", line 1137, in __call__
    return self.main(*args, **kwargs)
  File "/Users/timj/work/lsstsw3/miniconda/envs/lsst-scipipe-0.6.0/lib/python3.8/site-packages/click/core.py", line 1062, in main
    rv = self.invoke(ctx)
  File "/Users/timj/work/lsstsw3/miniconda/envs/lsst-scipipe-0.6.0/lib/python3.8/site-packages/click/core.py", line 1668, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/timj/work/lsstsw3/miniconda/envs/lsst-scipipe-0.6.0/lib/python3.8/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/timj/work/lsstsw3/miniconda/envs/lsst-scipipe-0.6.0/lib/python3.8/site-packages/click/core.py", line 763, in invoke
    return __callback(*args, **kwargs)
  File "/Users/timj/work/lsstsw3/build/daf_butler_test/python/lsst/daf/butler/cli/cmd/commands.py", line 117, in create
    script.createRepo(*args, **kwargs)
  File "/Users/timj/work/lsstsw3/build/daf_butler_test/python/lsst/daf/butler/script/createRepo.py", line 48, in createRepo
    raise RuntimeError("Oops")
RuntimeError: Oops
(lsst-scipipe-0.6.0) thrylos:daf_butler_test (tickets/DM-31696) $ butler create REPO
Traceback (most recent call last):
  File "/Users/timj/work/lsstsw3/build/daf_butler_test/python/lsst/daf/butler/cli/cmd/commands.py", line 119, in create
    script.createRepo(*args, **kwargs)
  File "/Users/timj/work/lsstsw3/build/daf_butler_test/python/lsst/daf/butler/script/createRepo.py", line 48, in createRepo
    raise RuntimeError("Oops")
RuntimeError: Oops
```
So yes, it does strip the "we don't care about this" code from the exception traceback. I assume we should fix bps and butler commands the same way. I had a feeling that @n8pease had done some exception catching trickery before but I can't found it now so maybe I was imagining it.

timj (approved):

python/lsst/ctrl/mpexec/cli/cmd/commands.py
andy-slac: I can update butler cli to do the same on this ticket if you prefer (and probably bps, but I never looked at that code).

timj: Thanks. BPS looks exactly like pipetask in that it uses click and is set up in the same way.

ktlim: Can't this instead be handled at a level higher at https://github.com/lsst/daf_butler/blob/master/python/lsst/daf/butler/cli/butler.py#L334 by stripping levels from the top instead of exiting out at the bottom?  That would mean one change instead of a change for every subcommand.

andy-slac: You would need to know how many stack levels you need to strip, which is sort of non-trivial. I am trying now to simplify things by defining a re-usable decorator which needs just one line change for each method  (but still need to update every method).

ktlim: I was thinking it would be relatively easy to strip until you get beyond `click`.

timj: Exactly. I had a similar problem in some astropy code working out what stacklevel to use for warning message to make the message show it came from user code (so I walked the stack). We know that click is at the top level and finding where click stops seems like a reasonable thing to do.

andy-slac: That feels a bit fragile as it depends on assumption that click does not call any other library before it lands in our code. Let me show you my decorator code and then you decide if that is too complicated.

andy-slac: OK, here is how I'm suggesting to do it [with a decorator](https://github.com/lsst/daf_butler/pull/572/commits/7681e57d883b660b8b1b04db5731309a6282ef79)

timj: Is it not possible to put that decorator once in the `main` definition that @ktlim pointed to above?

andy-slac: That will show click in traceback, to exclude click I'd need to do something much more non-trivial and fragile.

andy-slac: OK, another [attempt](https://github.com/lsst/daf_butler/pull/572/commits/906c6aa405397e3427f4310b830615d7eb4b5863) that does not require a per-method decorators, this depends on our overriding of `click.Command` so I can (ab)use it to wrap a callback method with the same decorator. This does not always work perfectly but in those cases I can keep explicit decorator (this only concerns three methods in ctrl_mpexec).

PR #141, opened 2021-09-16
DM-31583: Refresh registry caches before running a quantum.

TallJimbo:
This is necessary to pick up datasets added to the registry by other processes.  It may temporarily exacerbate problems with connection contention in shared-registry batch, but execution butler should fix that.

## Checklist

- [x] ran Jenkins


TallJimbo:
Closing this PR as I've (hopefully, pending Jenkins confirmation) fixed the problem another way down in daf_butler.

PR #143, opened 2021-09-28
DM-28111: minor improvements to dot file generation.

TallJimbo:
## Checklist

- [ ] ran Jenkins


python/lsst/ctrl/mpexec/dotTools.py
```diff
@@ -57,26 +58,36 @@ def _renderNode(file, nodeName, style, labels):
 
 def _renderTaskNode(nodeName, taskDef, file, idx=None):
     """Render GV node for a task"""
-    labels = [taskDef.taskName.rpartition('.')[-1]]
+    labels = [f"{taskDef.label}", taskDef.taskName]
```
andy-slac, line 61: I think we guarantee that task label is a string now, f-string should not be needed for it (and this is how it is done in _renderQuantumNode too)

andy-slac (approved):
Looks good, one minor comment.

PR #144, opened 2021-09-28
DM-31941: Do not purge datasets when clobbering in execution butler

timj:
It's much easier for provenance tracking if the UUID of the dataset always matches the one generated during initial creation of the execution butler. When clobbering the output datasets because of a resubmission of a failed job the pruneDatasets call was purging the records and deleting the files. This results in entirely new UUIDs being used when the new dataset
is put in the butler.  Modify the logic to not purge if it seems that an execution butler is being used.

## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/ctrl/mpexec/singleQuantumExecutor.py
+++ b/python/lsst/ctrl/mpexec/singleQuantumExecutor.py
@@ -327,7 +327,16 @@ class SingleQuantumExecutor(QuantumExecutor):
                 if self.clobberOutputs:
                     # only prune
                     _LOG.info("Removing partial outputs for task %s: %s", taskDef, existingRefs)
-                    butler.pruneDatasets(existingRefs, disassociate=True, unstore=True, purge=True)
+                    # Do not purge registry records if this looks like
+                    # an execution butler. This ensures that the UUID
+                    # of the dataset doesn't change.
+                    if butler._allow_put_of_predefined_dataset:
+                        purge = False
+                        disassociate = False
+                    else:
+                        purge = True
+                        disassociate = True
+                    butler.pruneDatasets(existingRefs, disassociate=disassociate, unstore=True, purge=purge)
                     return False
                 else:
                     raise RuntimeError(f"Registry inconsistency while checking for existing outputs:"


python/lsst/ctrl/mpexec/singleQuantumExecutor.py
```diff
@@ -327,7 +327,17 @@ def findOutputs(collections):
...
+                    # Do not purge registry records if this looks like
+                    # an execution butler. This ensures that the UUID
+                    # of the dataset doesn't change.
+                    if butler._allow_put_of_predefined_dataset:
+                        purge = False
+                        disassociate = False
+                    else:
+                        purge = True
+                        disassociate = True
+                    purge = False if butler._allow_put_of_predefined_dataset else True
```
ktlim, line 339: Wait, isn't this redundant with the above?

python/lsst/ctrl/mpexec/singleQuantumExecutor.py
```diff
@@ -327,7 +327,17 @@ def findOutputs(collections):
                 if self.clobberOutputs:
                     # only prune
                     _LOG.info("Removing partial outputs for task %s: %s", taskDef, existingRefs)
-                    butler.pruneDatasets(existingRefs, disassociate=True, unstore=True, purge=True)
+                    # Do not purge registry records if this looks like
+                    # an execution butler. This ensures that the UUID
+                    # of the dataset doesn't change.
+                    if butler._allow_put_of_predefined_dataset:
```
ktlim, line 333: Use of `_` suggests that maybe a more explicit API to determine execution butler may be warranted?

andy-slac (approved):
Looks good, I have the same comments as K-T.

python/lsst/ctrl/mpexec/singleQuantumExecutor.py
andy-slac: I agree with K-T, if we start depending on execution butler behavior outside daf_butler maybe it's time to add public method/property?

python/lsst/ctrl/mpexec/singleQuantumExecutor.py
timj: Yes. When testing I realized I needed to set the disassociate flag and left this line in by mistake.

python/lsst/ctrl/mpexec/singleQuantumExecutor.py
timj: Yes, that's the problem, but on the other hand ctrl_mpexec is not a million miles away from having to worry about the internals. I definitely don't want this to be a user visible property.

timj: @TallJimbo is hoping that execution butler is a transient concept and will disappear next year.

PR #160, opened 2022-01-14
DM-33278: Check for dataset type compatibility before failing

timj:
If the dataset types differ, it might be that the storage classes are different. If the storage classes are compatible with each other continue.

Requires lsst/daf_butler#632.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/ctrl/mpexec/preExecInit.py
+++ b/python/lsst/ctrl/mpexec/preExecInit.py
@@ -150,9 +150,17 @@ class PreExecInit:
                         "passing `--register-dataset-types` option to `pipetask run`."
                     ) from None
                 if expected != datasetType:
-                    raise ValueError(
-                        f"DatasetType configuration does not match Registry: {datasetType} != {expected}"
-                    )
+                    if expected.is_compatible_with(datasetType):
+                        _LOG.info(
+                            "The dataset type configurations differ (%s != %s from registry) "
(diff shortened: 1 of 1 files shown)


python/lsst/ctrl/mpexec/preExecInit.py
```diff
@@ -150,9 +150,17 @@ def initializeDatasetTypes(self, graph, registerDatasetTypes=False):
                         "passing `--register-dataset-types` option to `pipetask run`."
                     ) from None
                 if expected != datasetType:
-                    raise ValueError(
-                        f"DatasetType configuration does not match Registry: {datasetType} != {expected}"
-                    )
+                    if expected.is_compatible_with(datasetType):
+                        _LOG.info(
```
andy-slac, line 154: Should it be debug instead of info?

andy-slac (approved):
Looks good, one question.

timj:
I think the logic in this PR is not really right and needs to be redone.

* Firstly it clearly needs to distinguish get compatibility from put compatbility and so has to know if these dataset types are output or input (I have to look up what intermediates means).
* Secondly if we are registering the dataset type then the code is assuming that it's okay to always try to register but that will fail if they differ so an extra check has to go in there as well).

timj:
After talking about this on Slack I've decided to close this PR and come back to it with a more holistic approach that involves other parts of the system.

PR #161, opened 2022-01-20
DM-33303: Tests for pipelines where Task and Butler disagree on DatasetType

timj:
Depends on lsst/daf_butler#635 and lsst/pipe_base#228

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/ctrl/mpexec/preExecInit.py
+++ b/python/lsst/ctrl/mpexec/preExecInit.py
@@ -21,8 +21,6 @@
 
 __all__ = ["PreExecInit"]
 
-import itertools
-
 # -------------------------------
 #  Imports of standard modules --
 # -------------------------------
@@ -32,6 +30,7 @@ import logging
 #  Imports for other modules --
 # -----------------------------
 from lsst.base import Packages
+from lsst.daf.butler.registry import ConflictingDefinitionError
 from lsst.pipe.base import PipelineDatasetTypes
 
 _LOG = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


python/lsst/ctrl/mpexec/preExecInit.py
```diff
@@ -125,19 +124,55 @@ def initializeDatasetTypes(self, graph, registerDatasetTypes=False):
...
-            datasetTypes.initOutputs,
-            datasetTypes.intermediates,
-            datasetTypes.outputs,
+
+        for datasetTypes, is_input, is_output in (
+            (datasetTypes.initIntermediates, True, True),
+            (datasetTypes.initOutputs, False, True),
+            (datasetTypes.intermediates, True, True),
+            (datasetTypes.outputs, False, True),
         ):
```
andy-slac, line 133: `is_output` is always `True` here, maybe remove it from loop variables? Or did you also want to check pure inputs here?

python/lsst/ctrl/mpexec/preExecInit.py
```diff
@@ -125,19 +124,55 @@ def initializeDatasetTypes(self, graph, registerDatasetTypes=False):
...
+            if is_input:
+                # This DatasetType must be compatible on get.
+                is_compatible = datasetType.is_compatible_with(expected)
+            if is_output:
+                # Has to be able to be converted to expect type on put.
+                put_is_compatible = expected.is_compatible_with(datasetType)
+                if is_input:
+                    is_compatible = put_is_compatible and is_compatible
+                else:
+                    is_compatible = put_is_compatible
```
andy-slac, line 148: This looks more complicated than it should be. Maybe set `is_compatible` to `True` initially? Or is there a case when both `is_input` and `is_output` are false?

tests/test_simple_pipeline_executor.py
```diff
@@ -59,27 +81,150 @@ def test_from_task_class(self):
...
+            lsst.daf.butler.DatasetType(
+                "output",
+                dimensions=self.butler.registry.dimensions.empty,
+                storageClass="StructuredDataList",
+            )
+        )
+
+        with self.assertRaises(ValueError) as cm:
+            executor.run(register_dataset_types=True)
+        self.assertIn("inconsistent with registry definition", str(cm.exception))
```
andy-slac, line 226: Maybe `assertRaisesRegex` instead of separate test for exception message?

andy-slac (approved):
Looks good, couple of minor comments.

python/lsst/ctrl/mpexec/preExecInit.py
timj: Incredibly, I had not spotted that the `is_output` flag was always true. I had assumed when writing `_register_dataset_types` that there would be cases where input was true but outputs was false but seemingly I'm not making use of that. I suppose I should simplify the code below.

tests/test_simple_pipeline_executor.py
timj: Thanks. Had forgotten about that.

python/lsst/ctrl/mpexec/preExecInit.py
timj: Cleaned it up.

PR #171, opened 2022-02-21
DM-33638: Use datastore.mexists when checking for dataset existence

timj:
This can be much more efficient if there are many datasets
to be checked and a remote object store is being used.

## Checklist

- [x] ran Jenkins


andy-slac (approved):
Looks good.

PR #172, opened 2022-03-01
DM-33481: Add support for jobreport

andy-slac:
Added couple of pydantic data structures that contain data to save in
jobreport.json. `pipetask` adds a `--summary` option specifyng file
name for jobreport. Naturally pipetask cannot produce complete report,
in particular `exitCode` cannot be filled by the process before it
exits. Some post-processing of the generated file will be needed to
finalize it.

## Checklist

- [X] ran Jenkins

### DIFF
--- a/python/lsst/ctrl/mpexec/__init__.py
+++ b/python/lsst/ctrl/mpexec/__init__.py
@@ -25,6 +25,7 @@ from .executionGraphFixup import *
 from .mpGraphExecutor import *
 from .preExecInit import *
 from .quantumGraphExecutor import *
+from .reports import *
 from .simple_pipeline_executor import *
 from .singleQuantumExecutor import *
 from .taskFactory import *
--- a/python/lsst/ctrl/mpexec/cli/opt/options.py
+++ b/python/lsst/ctrl/mpexec/cli/opt/options.py
@@ -446,3 +446,12 @@ dataset_query_constraint = MWOptionDecorator(
     ),
     default="all",
 )
+
+summary_option = MWOptionDecorator(
+    "--summary",
+    help=(
+        "Location for storing job summary (JSON file). Note that the"
+        " structure of this file may not be stable."
+    ),
+    type=MWPath(dir_okay=False, file_okay=True, writable=True),
+)
--- a/python/lsst/ctrl/mpexec/mpGraphExecutor.py
+++ b/python/lsst/ctrl/mpexec/mpGraphExecutor.py
@@ -28,11 +28,8 @@ import pickle
 import signal
 import sys
 import time
-
-# -------------------------------
-#  Imports of standard modules --
-# -------------------------------
 from enum import Enum
+from typing import Optional
 
 from lsst.base import disableImplicitThreading
 from lsst.daf.butler.cli.cliLog import CliLog
@@ -43,6 +40,7 @@ from lsst.pipe.base.graph.graph import QuantumGraph
 #  Imports for other modules --
 # -----------------------------
 from .quantumGraphExecutor import QuantumGraphExecutor
+from .reports import ExecutionStatus, QuantumReport, Report
 
 _LOG = logging.getLogger(__name__)
 
@@ -71,6 +69,7 @@ class _Job:
         self.process = None
         self._state = JobState.PENDING
         self.started = None
+        self._rcv_conn = None
 
     @property
     def state(self):
@@ -93,11 +92,12 @@ class _Job:
         # it is pickled manually here.
         quantum_pickle = pickle.dumps(self.qnode.quantum)
         taskDef = self.qnode.taskDef
+        self._rcv_conn, snd_conn = multiprocessing.Pipe(False)
         logConfigState = CliLog.configState
         mp_ctx = multiprocessing.get_context(startMethod)
         self.process = mp_ctx.Process(
             target=_Job._executeJob,
-            args=(quantumExecutor, taskDef, quantum_pickle, butler, logConfigState),
+            args=(quantumExecutor, taskDef, quantum_pickle, butler, logConfigState, snd_conn),
             name=f"task-{self.qnode.quantum.dataId}",
         )
         self.process.start()
@@ -105,7 +105,7 @@ class _Job:
         self._state = JobState.RUNNING
 
     @staticmethod
-    def _executeJob(quantumExecutor, taskDef, quantum_pickle, butler, logConfigState):
+    def _executeJob(quantumExecutor, taskDef, quantum_pickle, butler, logConfigState, snd_conn):
         """Execute a job with arguments.
 
         Parameters
@@ -118,6 +118,8 @@ class _Job:
             Quantum for this task execution in pickled form.
         butler : `lss.daf.butler.Butler`
             Data butler instance.
+        snd_conn : `multiprocessing.Connection`
+            Connection to send job report to parent process.
         """
         if logConfigState and not CliLog.configState:
             # means that we are in a new spawned Python process and we have to
@@ -129,7 +131,15 @@ class _Job:
             butler.registry.resetConnectionPool()
 
         quantum = pickle.loads(quantum_pickle)
-        quantumExecutor.execute(taskDef, quantum, butler)
+        try:
+            quantumExecutor.execute(taskDef, quantum, butler)
+        finally:
+            # If sending fails we do not want this new exception to be exposed.
+            try:
+                report = quantumExecutor.getReport()
+                snd_conn.send(report)
+            except Exception:
+                pass
 
     def stop(self):
         """Stop the process."""
@@ -150,6 +160,23 @@ class _Job:
         if self.process and not self.process.is_alive():
             self.process.close()
             self.process = None
+            self._rcv_conn = None
+
+    def report(self) -> QuantumReport:
+        """Return task report, should be called after process finishes and
+        before cleanup().
+        """
+        try:
+            report = self._rcv_conn.recv()
+            report.exitCode = self.process.exitcode
+        except Exception:
+            # Likely due to the process killed, but there may be other reasons.
+            report = QuantumReport.from_exit_code(
+                exitCode=self.process.exitcode,
+                dataId=self.qnode.quantum.dataId,
+                taskLabel=self.qnode.taskDef.label,
+            )
+        return report
 
     def failMessage(self):
         """Return a message describing task failure"""
@@ -297,13 +324,21 @@ class MPGraphExecutor(QuantumGraphExecutor):
     """
 
     def __init__(
-        self, numProc, timeout, quantumExecutor, *, startMethod=None, failFast=False, executionGraphFixup=None
+        self,
+        numProc,
+        timeout,
+        quantumExecutor,
+        *,
+        startMethod=None,
+        failFast=False,
+        executionGraphFixup=None,
     ):
         self.numProc = numProc
         self.timeout = timeout
         self.quantumExecutor = quantumExecutor
         self.failFast = failFast
         self.executionGraphFixup = executionGraphFixup
+        self.report: Optional[Report] = None
 
         # We set default start method as spawn for MacOS and fork for Linux;
         # None for all other platforms to use multiprocessing default.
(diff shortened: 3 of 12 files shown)


andy-slac:
@timj, could you look at [pydantic classes](https://github.com/lsst/ctrl_mpexec/blob/97418a217948927ee4efd785cc267aa52c18f356/python/lsst/ctrl/mpexec/job_report.py), this is my first time doing something with pydantic, I want to make sure it is done right.

python/lsst/ctrl/mpexec/job_report.py
```diff
@@ -0,0 +1,187 @@
...
+    """Name of the exception class if exception was raised."""
+
+    message: str
+    """Exception message for in-process quantum execution, None if
+    quantum was executed in sub-process.
+    """
+
+    def __init__(self, exception: Exception):
+        """Construct instance from an exception."""
+        super().__init__(className=exception.__class__.__name__, message=str(exception))
```
timj, line 74: In theory we can use `lsst.utils.introspection.get_full_type_name()` -- depends on whether you want custom exceptions to include the full type name or just the short part.

python/lsst/ctrl/mpexec/job_report.py
```diff
@@ -0,0 +1,187 @@
...
+        self,
+        dataId: DataId,
+        taskLabel: str,
+        status: ExecutionStatus = ExecutionStatus.SUCCESS,
+        exitCode: Optional[int] = None,
+        exception: Optional[Exception] = None,
+    ):
+        super().__init__(
+            status=status.value,
+            dataId=_serializeDataId(dataId),
```
timj, line 107: The usual way of doing this sort of thing in pydantic is to define validators: https://pydantic-docs.helpmanual.io/usage/validators/ and https://github.com/lsst/daf_butler/blob/main/python/lsst/daf/butler/core/serverModels.py#L200-L212

You use those to examine a specific parameter and modify it if needed.

python/lsst/ctrl/mpexec/job_report.py
```diff
@@ -0,0 +1,187 @@
...
+            taskLabel=taskLabel,
+            exitCode=exitCode,
+            exceptionInfo=None if exception is None else ExceptionInfo(exception),
```
timj, line 110: Validator should work here.

python/lsst/ctrl/mpexec/job_report.py
```diff
@@ -0,0 +1,187 @@
...
+        cls,
+        exitCode: int,
+        dataId: DataId,
+        taskLabel: str,
+    ) -> QuantumReport:
+        """Construct report instance from an exit code and other pieces of
+        data.
+        """
+        return cls(
+            status=ExecutionStatus.SUCCESS if exitCode == 0 else ExecutionStatus.FAILURE,
```
timj, line 141: You might be able to have a validator derive status from the other parameters.

python/lsst/ctrl/mpexec/job_report.py
```diff
@@ -0,0 +1,187 @@
...
+        exception: Optional[Exception] = None,
+    ):
+        super().__init__(
+            status=status.value,
+            cmdLine=cmdLine if cmdLine is not None else _cmdLine(),
+            exitCode=exitCode,
+            exceptionInfo=None if exception is None else ExceptionInfo(exception),
+        )
+
+    def setException(self, exception: Exception) -> None:
```
timj, line 185: This is a new class so can we use `set_exception` and `from_exception` etc?

timj (approved):
I've had a quick look at the pydantic models. My main comment is to look at validators rather than custom `__init__`.

python/lsst/ctrl/mpexec/job_report.py
andy-slac: Somehow I cannot make it work, validators seem to be working OK for other cases, but fail for this one. Diagnostics is not helpful, I think I'm going to give up, and keep `__init__`, the code looks cleaner that way.

timj: Are you sure it's not that your `__init__` takes `exception` but your model expects `exceptionInfo`? Pydantic will pass the validator the exceptionInfo argument and allow you to change it. It doesn't like it if you pass in a completely different parameter.

andy-slac: I tried to do `QuantumReport(exceptionInfo=Exception, ...)` but that does not even get to validator, it is rejected earlier. I think it wants me to pass `ExceptionInfo` or something that it itself can convert into that class (it actually complains that `exceptionInfo` is not a `dict`).

python/lsst/ctrl/mpexec/job_report.py
```diff
@@ -0,0 +1,188 @@
...
+
+    Status `FAILURE` is set if one or more tasks failed. Status `TIMEOUT` is
+    set if there are no failures but one or more tasks timed out. Timeouts can
+    only be detected in multi-process mode, child task is killed on timeout
+    and usually should have non-zero exit code.
+    """
+
+    SUCCESS = "success"
+    FAILURE = "failure"
+    TIMEOUT = "timeout"
```
MichelleGower, line 59: When running pipelines_check, I made isr fail due by chmod'ing the raw file.  In the log, I got 

> INFO 2022-03-02T18:41:52.993-06:00 lsst.ctrl.mpexec.mpGraphExecutor ()(mpGraphExecutor.py:451) - Executed 0 quanta successfully, 1 failed and 2 remain out of total 3 quanta.
> 
> ERROR 2022-03-02T18:41:52.993-06:00 lsst.ctrl.mpexec.mpGraphExecutor ()(mpGraphExecutor.py:409) - Upstream job failed for task <TaskDef(lsst.pipe.tasks.characterizeImage.CharacterizeImageTask, label=characterizeImage) dataId={instrument: 'HSC', detector: 10, visit: 903342, ...}>, skipping this task.
> ERROR 2022-03-02T18:41:52.993-06:00 lsst.ctrl.mpexec.mpGraphExecutor ()(mpGraphExecutor.py:409) - Upstream job failed for task <TaskDef(lsst.pipe.tasks.calibrate.CalibrateTask, label=calibrate) dataId={instrument: 'HSC', detector: 10, visit: 903342, ...}>, skipping this task.

And in the jobreport.json file I got entries for all 3 quanta.  The report for isr made sense.  The reports for characterizeImage and calibrate said `"status": "failure",`  Were these downstream ones supposed to be included in the report?  If yes, can we have a different status (e.g., pruned).   Also should we have a different status for a quantum that is skipped because previously executed successfully?

python/lsst/ctrl/mpexec/cli/opt/options.py
```diff
@@ -446,3 +446,9 @@ def _to_int(value):
     ),
     default="all",
 )
+
+job_report_option = MWOptionDecorator(
+    "--job-report",
+    help="Location for storing job report (JSON file).",
```
MichelleGower, line 452: Even though this was requested for running PanDA jobs, I think there will be folks who start looking at these reports as soon as they discover the command line option.  Is there a standard way to warn that the contents in the file are still in flux?  (And if we're going to change the "job" to something else it would be nice to do it before these folks find this option.)

MichelleGower (approved):
I have an overall question about whether "job" should be changed to something more pipetask-y like executor.  And I'm not sure the behavior I saw with a failed quantum was what is expected.  And a few other misc items.  It would be nice if changing the "job" to get that done before in a stack, but if the "pruned"/skipped statuses need to wait until another ticket that's fine.

python/lsst/ctrl/mpexec/mpGraphExecutor.py
```diff
@@ -129,7 +131,16 @@ def _executeJob(quantumExecutor, taskDef, quantum_pickle, butler, logConfigState
             butler.registry.resetConnectionPool()
 
         quantum = pickle.loads(quantum_pickle)
-        quantumExecutor.execute(taskDef, quantum, butler)
+        try:
+            quantumExecutor.execute(taskDef, quantum, butler)
+            report = QuantumReport(dataId=quantum.dataId, taskLabel=taskDef.label)
```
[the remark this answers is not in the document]
andy-slac: Sorry, this was a leftover from an earlier version when `quantumExecutor` did not have `getReport` method, I will remove that.

python/lsst/ctrl/mpexec/__init__.py
```diff
@@ -22,6 +22,7 @@
 from .cmdLineFwk import *
 from .dotTools import *
 from .executionGraphFixup import *
+from .job_report import *
```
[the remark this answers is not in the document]
andy-slac: I could rename module to `reports` and `JobReport` class to `QuantumGraphReport` to keep symmetry between QuantumExecutor/QuantumReport and QuantumGraphExecutor/QuantumGraphReport? 

I think Tim's idea was that `JobReport/QuantumGraphReport` structure would be used by bps/pilot as well, e.g. you could read pipetask-made JSON file into that structure, update it with more info and save it again. But if `QuantumGraphReport` is going to be pipetask-specific I can also drop `exitMsg` from that struct as I do not know what to put there.

👍 for `cmdLine`.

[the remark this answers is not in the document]
timj: Yes. We need to decide whether pipetask is writing a subset of the job report content and then something else is coming in and attaching this subset to a full report that panda would need, or if this is a representation of all the information we want PanDA to get and so it's the full model with some optional bits that need to be filled in later.

python/lsst/ctrl/mpexec/mpGraphExecutor.py
```diff
@@ -451,11 +498,15 @@ def _executeQuantaMP(self, graph, butler):
                     _LOG.debug("finished: %s", job)
                     # finished
                     exitcode = job.process.exitcode
+                    report = job.report()
+                    if report:
+                        self.jobReport.quantaReports.append(report)
```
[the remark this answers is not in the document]
andy-slac: I see that it can be confusing in this context. The Job class in this module is not exposed, it's just an implementation detail of this executor. JobReport class is what appears in QuantumGraphExecutor interface, so I don't think there will be a confusion about what it means.

python/lsst/ctrl/mpexec/job_report.py
```diff
@@ -0,0 +1,188 @@
...
+    """
+
+    exceptionInfo: Optional[ExceptionInfo] = None
+    """Exception information if exception was raised."""
+
+    def __init__(
+        self,
+        dataId: DataId,
+        taskLabel: str,
+        status: ExecutionStatus = ExecutionStatus.SUCCESS,
```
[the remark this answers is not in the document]
andy-slac: I don't think there is a risk, we have special class methods to create reports with FAILURE status, should be reasonably OK.

tests/test_executors.py
```diff
@@ -50,15 +57,28 @@ def __init__(self, mp=False):
...
+                    exception=exc,
+                    dataId=quantum.dataId,
+                    taskLabel=taskDef.label,
+                )
+                raise
         self.quanta.append(quantum)
         return quantum
 
+    def getReport(self):
+        return self.report
```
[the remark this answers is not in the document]
andy-slac: Right, because this is not the real thing I thought there is no point in trying to test the exception, but I will reconsider.

python/lsst/ctrl/mpexec/job_report.py
andy-slac: I think I decided not to include skipped quanta, but looking at the code now they are included for in-process running (I guess you ran above test with -j1?) In multi-process they should not appear. I have no big preference, if you prefer I can add "pruned" status for those and always include them.

python/lsst/ctrl/mpexec/cli/opt/options.py
andy-slac: I can add a warning to the help string, I don't want to issue an actual WARNING for it. Let's try to decide on a name during the meeting.

python/lsst/ctrl/mpexec/__init__.py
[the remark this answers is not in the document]
andy-slac: I have renamed `job_report` module to `reports` and `JobReport` class to just `Report`. `exitMsg` dropped from the `Reporty` class, we can re-add things later if needed.

python/lsst/ctrl/mpexec/job_report.py
andy-slac: I added SKIPPED="skipped" status, setting it for all skipped dependencies.

python/lsst/ctrl/mpexec/cli/opt/options.py
andy-slac: Command line option was renamed to `--summary`.

timj:
Quick question: is there a test that Pydantic can read the JSON back in? (`model.parse_obj()`) I'm wondering about the disconnect between the constructor parameters and the model names -- pydantic usually does not like those to be different.

andy-slac:
> Quick question: is there a test that Pydantic can read the JSON back in? (`model.parse_obj()`) I'm wondering about the disconnect between the constructor parameters and the model names -- pydantic usually does not like those to be different.

I did not check that, will add a unit test to see if it works.

andy-slac:
Indeed, custom `__init__` did not work very well with pydantic models. I had to refactor the classes again, but now things work OK and JSON round-trip is working too.

PR #174, opened 2022-03-16
DM-34076: Improve timeout handling by MPGraphExecutor

andy-slac:
Detect the case when a task manages to finish successfully before it
gets killed due to timeout. This guarantees more consistent status in
a quantum report.

Increase running time for timed out task in a unit test. 
Running time for timing-out task increased to 100 seconds, this should
reduce the chance for it to finish successfully before timeout is
detected by a parent.

## Checklist

- [X] ran Jenkins

### DIFF
--- a/python/lsst/ctrl/mpexec/mpGraphExecutor.py
+++ b/python/lsst/ctrl/mpexec/mpGraphExecutor.py
@@ -70,11 +70,18 @@ class _Job:
         self._state = JobState.PENDING
         self.started = None
         self._rcv_conn = None
+        self._terminated = False
 
     @property
     def state(self):
         """Job processing state (JobState)"""
         return self._state
+
+    @property
+    def terminated(self):
+        """Return True if job was killed by stop() method and negative exit
+        code is returned from child process. (`bool`)"""
+        return self._terminated and self.process.exitcode < 0
 
     def start(self, butler, quantumExecutor, startMethod=None):
         """Start process which runs the task.
@@ -152,6 +159,7 @@ class _Job:
         else:
             _LOG.debug("Killing process %s", self.process.name)
             self.process.kill()
+        self._terminated = True
 
     def cleanup(self):
         """Release processes resources, has to be called for each finished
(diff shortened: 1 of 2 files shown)


python/lsst/ctrl/mpexec/mpGraphExecutor.py
```diff
@@ -70,12 +70,18 @@ def __init__(self, qnode):
...
+        self._terminated = False
 
     @property
     def state(self):
         """Job processing state (JobState)"""
         return self._state
 
+    @property
+    def terminated(self):
+        """Return True if job was killed by stop() method (`bool`)"""
```
timj, line 82: but also if the exit code was negative?

python/lsst/ctrl/mpexec/mpGraphExecutor.py
```diff
@@ -501,51 +511,48 @@ def _executeQuantaMP(self, graph, butler):
...
                         job.cleanup()
                         _LOG.debug("success: %s took %.3f seconds", job, time.time() - job.started)
                     else:
-                        self.report.status = ExecutionStatus.FAILURE
-                        # failMessage() has to be called before cleanup()
-                        message = job.failMessage()
-                        jobs.setJobState(job, JobState.FAILED)
+                        if job.terminated:
+                            # Was killed due to timeout.
+                            if self.report.status == ExecutionStatus.SUCCESS:
```
timj, line 522: If it is killed due to time out and reports success why can't that be treated as success? How can you be timed out and successful?

timj (approved):
This looks okay. My main confusion is why something that is timed out but still finishes before the time out completes can't be treated as a success.

python/lsst/ctrl/mpexec/mpGraphExecutor.py
andy-slac: This is sort of implied, if `stop` was called but did not kill the process then it was not killed by stop. I'll expand the docstring.

python/lsst/ctrl/mpexec/mpGraphExecutor.py
andy-slac: It is treated as success, `terminated` is False for that case, and check for `exitcode == 0` is in the `if` branch above. This branch is only for the case when it was actually killed by a signal (had non-zero `exitcode`)

PR #177, opened 2022-03-28
DM-34215: Support dropping into debugger on exception

PaulPrice:
When we get an exception, sometimes it's not enough to just
see the backtrace, but we also want to get a debugger prompt
at the point of the exception.

Replaced the --do-raise option (apparently unused, and often
associated with getting a debugger with Gen2) with --pdb
(well-known and appreciated option from pytest), which causes
MPGraphExecutor._executeQuantaInProcess to drop into pdb.

## Checklist

- [ ] ran Jenkins

### DIFF
--- a/doc/changes/DM-34215.feature.rst
+++ b/doc/changes/DM-34215.feature.rst
@@ -1,0 +1,4 @@
+Replaced the unused ``--do-raise`` option with ``--pdb``,
+which drops the user into the debugger
+(``pdb`` by default, but ``--pdb=ipdb`` also works if you have ``ipdb`` installed)
+on an exception.
--- a/python/lsst/ctrl/mpexec/cli/opt/options.py
+++ b/python/lsst/ctrl/mpexec/cli/opt/options.py
@@ -76,8 +76,12 @@ delete_option = MWOptionDecorator(
 )
 
 
-do_raise_option = MWOptionDecorator(
-    "--do-raise", help="Raise an exception on error. (else log a message and continue?)", is_flag=True
+pdb_option = MWOptionDecorator(
+    "--pdb",
+    help="Post-mortem debugger to launch for exceptions (defaults to pdb if unspecified; requires a tty).",
+    is_flag=False,
+    flag_value="pdb",
+    default=None,
 )
 
 
--- a/python/lsst/ctrl/mpexec/mpGraphExecutor.py
+++ b/python/lsst/ctrl/mpexec/mpGraphExecutor.py
@@ -22,6 +22,7 @@
 __all__ = ["MPGraphExecutor", "MPGraphExecutorError", "MPTimeoutError"]
 
 import gc
+import importlib
 import logging
 import multiprocessing
 import pickle
@@ -330,6 +331,9 @@ class MPGraphExecutor(QuantumGraphExecutor):
         one for current platform.
     failFast : `bool`, optional
         If set to ``True`` then stop processing on first error from any task.
+    pdb : `str`, optional
+        Debugger to import and use (via the ``post_mortem`` function) in the
+        event of an exception.
     executionGraphFixup : `ExecutionGraphFixup`, optional
         Instance used for modification of execution graph.
     """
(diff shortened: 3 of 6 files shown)


doc/changes/DM-34215.feature.rst
```diff
@@ -0,0 +1 @@
+Replaced the unused `--do-raise` option with `--pdb`, which drops the user into the debugger on an exception.
```
timj, line 1: Needs double back ticks around these for RST.

python/lsst/ctrl/mpexec/mpGraphExecutor.py
```diff
@@ -436,6 +438,10 @@ def _executeQuantaInProcess(self, graph, butler):
                 self.quantumExecutor.execute(qnode.taskDef, qnode.quantum, butler)
                 successCount += 1
             except Exception as exc:
+                if self.pdb and self.numProc == 1:
+                    import pdb  # noqa E402: only load this when we need it
```
mfisherlevine, line 442: Ooooh, while we're in here, any chance of changing this to be something I've wanted for ages?!

```
    try:
        import ipdb as pdb:
    except ImportError:
        import pdb
```

PaulPrice: What if `ipdb` is available, but I actually want `pdb`?
I think this would need some more configuration.

mfisherlevine: Oh, I guess technically that's possible, I admit I didn't think of that. I assumed that `ipdb` was pretty much a superset of pdb, and that if someone had gone to the trouble of installing it, that would mean it was their debugger of choice, but maybe neither of those assumptions are actually true, and sometimes you actually need strictly `pdb`.

PaulPrice: OK, changed to allow `--pdb=ipdb`.

mfisherlevine: Oh amazing, thank you!

python/lsst/ctrl/mpexec/cli/opt/options.py
```diff
@@ -76,8 +76,12 @@ def _to_int(value):
 )
 
 
-do_raise_option = MWOptionDecorator(
-    "--do-raise", help="Raise an exception on error. (else log a message and continue?)", is_flag=True
+pdb_option = MWOptionDecorator(
+    "--pdb",
+    help="Drop into pdb on exception (only if single process)?",
```
andy-slac, line 81: Do we need a question mark here? 
Can you also expand it a bit and say that it can specify different debugger, e.g. ipdb?

python/lsst/ctrl/mpexec/mpGraphExecutor.py
```diff
@@ -436,6 +442,24 @@ def _executeQuantaInProcess(self, graph, butler):
                 self.quantumExecutor.execute(qnode.taskDef, qnode.quantum, butler)
                 successCount += 1
             except Exception as exc:
+                if self.pdb and self.numProc == 1:
```
andy-slac, line 445: Check for `numProc` is probably not needed here, this code always runs in-process which means no parallelism.
I'd probably add a paranoid check that stdin/stdout are connected to a terminal (`sys.std{in,out}.isatty()`).

andy-slac (approved):
Looks good, couple of minor comments.

python/lsst/ctrl/mpexec/mpGraphExecutor.py
TallJimbo: Could we get open-ended debugger choice by just using `breakpoint()`, and trying the user to run inside their favorite debugger?  I believe that's what `breakpoint()` is for.

PaulPrice: `breakpoint()` is for getting a debugger at the current point of execution. However, we want to get a debugger at the point of execution where the exception was raised. It's the difference between `pdb.set_trace()` and `pdb.post_mortem()`.
We could have Merlin set `PYTHONBREAKPOINT=ipdb.post_mortem`, but I don't think that would work if he's sprinkled other `breakpoint()` calls throughout his code.

mfisherlevine: I don't think I'd tend to mix-and-match other breakpoints with this, but who knows. Also though, I know that both pdb and ipdb support post_mortem, so I'm not sure what other debuggers there are which wouldn't work with this. Basically, I think this is great as it is 🙂

python/lsst/ctrl/mpexec/mpGraphExecutor.py
PaulPrice: I like the tty check. That also catches the case where we're running with multiple processes, which was my main concern.

PR #190, opened 2022-06-10
DM-35047: Add upload to pypi

mwittgen:
## Checklist

- [ ] ran Jenkins

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,17 +1,24 @@
 [metadata]
-name = ctrl_mpexec
+name = lsst-ctrl-mpexec
 description = Pipeline execution infrastructure for the Rubin Science Pipelines
 author = Rubin Observatory Data Management
 url = https://github.com/lsst/resources
+author_email = dm-admin@lists.lsst.org
+license = GPLv3+ License
 classifiers =
     Intended Audience :: Science/Research
-    License :: OSI Approved ::  GNU General Public License v3 or later (GPLv3+)
+    License :: OSI Approved :: GNU General Public License v3 or later (GPLv3+)
     Operating System :: OS Independent
     Programming Language :: Python :: 3
     Programming Language :: Python :: 3.8
+    Programming Language :: Python :: 3.9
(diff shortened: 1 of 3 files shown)


setup.cfg
```diff
@@ -1,18 +1,25 @@
 [metadata]
-name = ctrl_mpexec
+name = lsst-ctrl-mpexec
 description = Pipeline execution infrastructure for the Rubin Science Pipelines
 author = Rubin Observatory Data Management
 url = https://github.com/lsst/resources
```
timj, line 5: Wrong URL

setup.cfg
```diff
@@ -1,18 +1,25 @@
...
+author_email = dm-admin@lists.lsst.org
+license = GPLv3+ License
 classifiers =
     Intended Audience :: Science/Research
-    License :: OSI Approved ::  GNU General Public License v3 or later (GPLv3+)
+    License :: OSI Approved :: GNU General Public License v3 or later (GPLv3+)
     Operating System :: OS Independent
     Programming Language :: Python :: 3
     Programming Language :: Python :: 3.8
```
timj, line 13: I thought this now failed on python 3.8?

timj (approved):

setup.cfg
mwittgen: So this should use solely python 3.10?

timj: I think you are replying to the other comment...

setup.cfg
mwittgen: So this should use 3.10?

timj: I thought you were testing the different python versions before declaring them to work. I think ctrl_mpexec has a problem with 3.8 type annotations (found by ctrl_bps). Maybe declare 3.10 then but in a follow up ticket we should enable 3.8, 9, 10 matrix and allow us to properly state which versions we support (we are not actively preventing 3.8 from working but if it stops working we do not care).

mwittgen: OK.

PR #191, opened 2022-06-10
DM-35191: Add test matrix for python versions

mwittgen:
## Checklist

- [ ] ran Jenkins

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -11,6 +11,10 @@ on:
 jobs:
   build_and_test:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ["3.9", "3.10"]
+
     steps:
       - uses: actions/checkout@v3
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yaml
```diff
@@ -77,7 +81,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v3
         with:
-          python-version: "3.10"
+          python-version: "3.9"
```
timj, line 84: Why is this now 3.9?

timj (approved):
One question.

.github/workflows/build.yaml
mwittgen: 3.9 and 3.10 pass the matrix test. Would we pick the lowest python version for publishing? Or just use 3.10 and 
also declare 3.10 required?

mwittgen: Changed it back to 3.10

PR #192, opened 2022-06-15
Use build dependencies from github

mwittgen:
## Checklist

- [ ] ran Jenkins

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -55,7 +55,7 @@ jobs:
       - name: Build and install
         shell: bash -l {0}
         run: |
-          pip install -v -e .
+          pip install -v --no-deps -e .
 
       - name: Run tests
         shell: bash -l {0}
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yaml
```diff
@@ -92,8 +92,7 @@ jobs:
 
       - name: Build and create distribution
         run: |
-          pip install -v -e .
-          python -m build --sdist --wheel
+          python -m build --skip-dependency-check --sdist --wheel
```
timj, line 95: Don't we need to do `pip install -r requirements.txt` before this? (I think in the other PRs as well)

mwittgen: It doesn't seem to be required and would have no effect. python-build creates its own venv 
https://github.com/mwittgen/daf_butler/runs/6889899166?check_suite_focus=true#step:6:1962
I need to better understand this.

mwittgen: When I write a non-sense package name into setup.cfg, python-build just proceeds, but the
METADATA in the wheel just takes whatever is in setup.cfg
Requires-Dist: lsst-sphgeom-broken

timj: Oh, you are saying that it's irrelevant because building the package to make a wheel doesn't require the dependencies to be imported and so it doesn't care if the dependencies are available or not?

mwittgen: Right

timj (approved):

PR #194, opened 2022-07-04
DM-35082 Use DimensionUniverse from QuantumGraph

natelust:
## Checklist

- [ ] ran Jenkins


andy-slac:
@natelust, it looks like there is a fallout from pex_config updates and mypy check is failing, could you check it?

timj:
It's likely that requirements.txt needs to be temporarily hacked to point at the pipe_base branch.

python/lsst/ctrl/mpexec/cmdLineFwk.py
```diff
@@ -570,9 +570,7 @@ def makeGraph(self, pipeline: Pipeline, args: SimpleNamespace) -> Optional[Quant
         if args.qgraph:
             # click passes empty tuple as default value for qgraph_node_id
             nodes = args.qgraph_node_id or None
-            qgraph = QuantumGraph.loadUri(
-                args.qgraph, butler.registry.dimensions, nodes=nodes, graphID=args.qgraph_id
-            )
+            qgraph = QuantumGraph.loadUri(args.qgraph, nodes=nodes, graphID=args.qgraph_id)
```
timj, line 573: Does it hurt to pass the dimension universe through in this call? If not, is ctrl_mpexec supposed to check that the universes are compatible itself?

PR #195, opened 2022-07-08
DM-35494 Allow execution butler creation to transfer datasets

PaulPrice:
## Checklist

- [ ] ran Jenkins


timj (changes requested):
I worry that BPS is going to want to be a bit cleverer for this and we should be including an option to specify the execution butler datastore root. Writing all the data into the place where we also write the execution butler could cause confusion for when jobs copy the execution butler to each node. It's also going to be a bit odd that the execution butler is copied locally and all the jobs write their outputs to the original execution butler location. Decoupling the location of the execution butler from the location of the execution datastore seems like it should be an option. Please add a `--execution-datastore-root` (or similar, bearing in mind that this option will have to exist for graph-backed butler to work for you cc/ @andy-slac ) option which can default to empty and imply it will go to the execution butler root if `--transfer` is set. It's a shame that `transfer=auto` can't trigger based on this datastore root override but then there would be no way to do what you want of copying the files into the execution butler tree.

python/lsst/ctrl/mpexec/cli/opt/optionGroups.py
```diff
@@ -88,6 +89,8 @@ def __init__(self) -> None:
             ctrlMpExecOpts.qgraph_dot_option(),
             ctrlMpExecOpts.save_execution_butler_option(),
             ctrlMpExecOpts.clobber_execution_butler_option(),
+            ctrlMpExecOpts.datastore_root_option(),
+            transfer_option(),
```
timj, line 93: Are you not wanting to override the help text?

python/lsst/ctrl/mpexec/cli/opt/options.py
```diff
@@ -415,6 +415,15 @@
     ),
     is_flag=True,
 )
+
+datastore_root_option = MWOptionDecorator(
+    "--datastore-root",
```
timj, line 420: `--target-datastore-root`?

python/lsst/ctrl/mpexec/cli/script/qgraph.py
```diff
@@ -135,6 +137,9 @@ def qgraph(  # type: ignore
         QuantumGraph.
     clobber_execution_butler : `bool`
         It True overwrite existing execution butler files if present.
+    transfer : `str`
+        Transfer mode for execution butler creation. The default option,
+        "none", means that no files will be transferred.
```
timj, line 142: Where is `none` coming in as an option? It's not the available in the standard list.  `auto` will not quite work if the source and destination are both POSIX because it will try symlink before it tries copy if the target and source roots differ. That might be fine though since you are likely going to want an explicit "copy" here with a new root. "direct" would leave all the existing files where they are but let you write new files to the new root.

You could make "auto" the default if no root is specified and "copy" the default if a new root is given.

timj (approved):

python/lsst/ctrl/mpexec/cli/script/qgraph.py
```diff
@@ -135,6 +137,12 @@ def qgraph(  # type: ignore
         QuantumGraph.
     clobber_execution_butler : `bool`
         It True overwrite existing execution butler files if present.
+    target_datastore_root : `str` or `None`
+        URI location for the execution butler's datastore.
+    transfer : `str` or `None`
```
timj, line 142: I don't think this is ever going to be `None` is it?

PR #199, opened 2022-08-20
DM-35917: Migrate config setting docs from pipe_base and update code to match expectations

timj:
Whilst migrating some config docs from pipe_base I realized that the --show option was not working right. I ended up trying to fix it.

## Checklist

- [x] ran Jenkins


python/lsst/ctrl/mpexec/cli/cmd/commands.py
```diff
@@ -126,8 +138,17 @@ def qgraph(ctx: click.Context, **kwargs: Any) -> None:
 def run(ctx: click.Context, **kwargs: Any) -> None:
     """Build and execute pipeline and quantum graph."""
     kwargs = _collectActions(ctx, **kwargs)
-    pipeline = script.build(**kwargs)
-    qgraph = script.qgraph(pipelineObj=pipeline, **kwargs)
+    show = ShowInfo(kwargs.pop("show", []))
+    pipeline = script.build(**kwargs, show=show)
+    if show.handled and not show.unhandled:
+        # The show option was given and all options were processed.
```
timj, line 144: I wonder if we should print this to the user so they aren't confused.

python/lsst/ctrl/mpexec/cli/script/qgraph.py
```diff
@@ -200,7 +199,6 @@ def qgraph(  # type: ignore
         raise RuntimeError("QuantumGraph is empty.")
 
     # optionally dump some info.
-    if show:
-        f.showInfo(args, pipelineObj, qgraph)
+    show.show_graph_info(qgraph, args)
```
andy-slac, line 202: Line 133 above - docstring needs an update for the new type of `show` parameter.

andy-slac (approved):
Looks great, few minor comments.

python/lsst/ctrl/mpexec/cli/cmd/commands.py
andy-slac: Hmm, if the output from e.g. `dump-config` is redirected to a file, it may also confuse something.

timj: We can  print to stderr?

doc/lsst.ctrl.mpexec/configuring-pipetask-tasks.rst
```diff
@@ -0,0 +1,365 @@
...
+- ``calibrate.astrometry`` has a subtask named ``matcher``, which is implemented by `~lsst.meas.astrom.matchOptimisticB.matchOptimisticBContinued.MatchOptimisticBTask`.
+
+Note that if the ``calibrate.astrometry`` task is retargeted to a different task class, the subtask of ``calibrate.astrometry`` *may* change (for example, ``calibrate.astrometry.matcher`` may no longer exist).
+
+.. _command-line-config-howto-config:
+
+How to set configurations with command-line arguments
+=====================================================
+
+Pipelines can be configured through a combination of two mechanisms: arguments on the command line (``--config``) or through configuration files (:option:``--config-file``).
```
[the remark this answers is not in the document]
timj: `:option:` is a pipe_base comand-line task thing that automatically links to the option description -- but for the click generated docstring I don't think we can link to specific options. I was meant to be removing them all.

python/lsst/ctrl/mpexec/cli/cmd/commands.py
andy-slac: stderr is probably better than stdout, though we should probably ask users about what they want. I myself almost never use `pipetask`, so my opinion probably should not count.

PR #204, opened 2022-09-27
DM-36360: Improve diagnostic visibility for empty QGs.

TallJimbo:
By moving the exception raise down to the outer click layer (returning None through all higher layers) we avoid the long traceback that hides the useful information being logged by pipe_base.

## Checklist

- [x] ran Jenkins
- [x] added a release note for user-visible changes to `doc/changes` (in pipe_base)

I expect the `build_and_test` check here to fail until the pipe_base is branch is merged.


parejkoj (approved):
Wording suggestion on the exception message, but otherwise this looks good.

PR #206, opened 2022-10-05
DM-36412: Try to set the datastore cache directory explicitly

timj:
## Checklist

Needs lsst/daf_butler#738

- [x] ran Jenkins

### DIFF
--- a/python/lsst/ctrl/mpexec/cmdLineFwk.py
+++ b/python/lsst/ctrl/mpexec/cmdLineFwk.py
@@ -26,18 +26,17 @@ from __future__ import annotations
 
 __all__ = ["CmdLineFwk"]
 
-# -------------------------------
-#  Imports of standard modules --
-# -------------------------------
+import atexit
 import copy
 import datetime
 import getpass
 import logging
+import shutil
 from collections.abc import Iterable, Sequence
 from types import SimpleNamespace
 from typing import Optional, Tuple
 
-from lsst.daf.butler import Butler, CollectionType, DatasetRef, Registry
+from lsst.daf.butler import Butler, CollectionType, DatasetRef, DatastoreCacheManager, Registry
 from lsst.daf.butler.registry import MissingCollectionError, RegistryDefaults
 from lsst.daf.butler.registry.wildcards import CollectionWildcard
 from lsst.pipe.base import (
@@ -317,6 +316,7 @@ class _ButlerFactory:
             A read-only butler initialized with the collections specified by
             ``args``.
         """
+        cls.defineDatastoreCache()  # Ensure that this butler can use a shared cache.
         butler, inputs, _ = cls._makeReadParts(args)
         _LOG.debug("Preparing butler to read from %s.", inputs)
         return Butler(butler=butler, collections=inputs)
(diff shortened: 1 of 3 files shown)


python/lsst/ctrl/mpexec/cmdLineFwk.py
```diff
@@ -351,6 +353,30 @@ def makeButlerAndCollections(cls, args: SimpleNamespace) -> Tuple[Butler, Sequen
...
+        """
+        if os.environ.get("DAF_BUTLER_CACHE_DIRECTORY_IF_UNSET"):
+            # A value has already been set.
+            return
+
+        # Come up with a cache name. Create it and immediately delete
+        # it -- there is no nice way to determine a name without creating
+        # it (understandably, this can cause a race condition).
+        cache_dir = tempfile.mkdtemp(prefix="ctrl-mpexec-datastore-cache-")
+        os.rmdir(cache_dir)
```
andy-slac, line 375: What happens if you don't remove it here?

python/lsst/ctrl/mpexec/cmdLineFwk.py
```diff
@@ -351,6 +353,30 @@ def makeButlerAndCollections(cls, args: SimpleNamespace) -> Tuple[Butler, Sequen
...
+        """Define where datastore cache directories should be found.
+
+        Notes
+        -----
+        All the jobs should share a datastore cache if applicable. This
+        method determines a name and sets an environment variable that
+        will be picked up by the caching class. A value will not be set
+        if one is already defined.
+        """
+        if os.environ.get("DAF_BUTLER_CACHE_DIRECTORY_IF_UNSET"):
```
andy-slac, line 367: Would it be better to hide this sort of details (envvar name) in `DatastoreCacheManager` class? Or maybe even the whole method?

andy-slac (approved):
Looks good.

python/lsst/ctrl/mpexec/cmdLineFwk.py
timj: It *should* be automatically deleted at the end regardless. The directory is created during processing only if a cache is created and currently a cache is created only if a file datastore is talking to an object store. Ideally I wouldn't create a directory at all since it won't be needed for a local filesystem datastore and it seems bad to always create it. mkdtemp doesn't let me just come up with a name, so the other option is to create the random name myself and not use mkdtemp at all.

python/lsst/ctrl/mpexec/cmdLineFwk.py
timj: I think I do still need to set an environment variable somewhere since that's the only thing we guarantee will end up in a spawned subprocess without having to put code in the executeJob method. The DatastoreCacheManager class is a bit of an internal detail but I'll see about adding a `DatastoreCacheManager.set_fallback_cache_dir()` method that still uses the environment.

python/lsst/ctrl/mpexec/cmdLineFwk.py
andy-slac: Well, I was thinking about uniqueness guarantee of `mkdtemp`. My understanding is that can provide that guarantee because it checks that directory is not already there. If you remove directory but continue to use its name, then another call to `mkdtemp` may create the same directory again?

timj: If there is a clash the main hilarity will be another `pipetask` deleting the cache directory whilst the other process is using it.  I have no means of asking whether a cache directory would be needed or not since it all depends on how that specific butler was configured -- the only way to really know that is an explicit datastore API. If I make my own temp name with 20 random letters in it it seems unlikely that there would ever really be a clash and I can check that it wasn't already in use.

timj: Okay. I've just pushed a change that includes the process ID and 16 character random string in the temp dir name and I'm hoping that this is going to have a fairly low chance of clashing.

PR #207, opened 2022-10-05
DM-36145: Add additional quanta information for pipetask run

leeskelvin:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/ctrl/mpexec/cmdLineFwk.py
+++ b/python/lsst/ctrl/mpexec/cmdLineFwk.py
@@ -36,6 +36,7 @@ from collections.abc import Iterable, Sequence
 from types import SimpleNamespace
 from typing import Optional, Tuple
 
+from astropy.table import Table
 from lsst.daf.butler import Butler, CollectionType, DatasetRef, DatastoreCacheManager, Registry
 from lsst.daf.butler.registry import MissingCollectionError, RegistryDefaults
 from lsst.daf.butler.registry.wildcards import CollectionWildcard
@@ -586,18 +587,20 @@ class CmdLineFwk:
             if args.show_qgraph_header:
                 qgraph.buildAndPrintHeader()
 
-        # Count quanta in graph and give a warning if it's empty and return
-        # None.
+        # Count quanta in graph; give a warning if it's empty and return None.
         nQuanta = len(qgraph)
         if nQuanta == 0:
             return None
         else:
-            _LOG.info(
-                "QuantumGraph contains %d quanta for %d tasks, graph ID: %r",
-                nQuanta,
-                len(qgraph.taskGraph),
-                qgraph.graphID,
-            )
+            if _LOG.isEnabledFor(logging.INFO):
+                qg_task_table = self._generateTaskTable(qgraph)
+                _LOG.info(
+                    "QuantumGraph contains %d quanta for %d tasks, graph ID: %r\n%s",
+                    nQuanta,
+                    len(qgraph.taskGraph),
+                    qgraph.graphID,
+                    str(qg_task_table),
+                )
 
         if args.save_qgraph:
             qgraph.saveUri(args.save_qgraph)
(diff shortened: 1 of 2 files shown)


python/lsst/ctrl/mpexec/cmdLineFwk.py
```diff
@@ -576,12 +576,18 @@ def makeGraph(self, pipeline: Pipeline, args: SimpleNamespace) -> Optional[Quant
         if nQuanta == 0:
             return None
         else:
-            _LOG.info(
-                "QuantumGraph contains %d quanta for %d tasks, graph ID: %r",
-                nQuanta,
-                len(qgraph.taskGraph),
-                qgraph.graphID,
+            qg_tasks = []
```
timj, line 579: Since iterating through the graph is not "free" can you add a `if _LOG.isEnabledFor(logging.INFO)`  around this?

python/lsst/ctrl/mpexec/cmdLineFwk.py
```diff
@@ -586,18 +586,24 @@ def makeGraph(self, pipeline: Pipeline, args: SimpleNamespace) -> Optional[Quant
...
+                qg_tasks = []
+                for task_def in qgraph.iterTaskGraph():
+                    qnodes = qgraph.getNodesForTask(task_def)
+                    qg_tasks.append(f"{len(qnodes):>8} {task_def.label}")
+                task_col_max_len = len(max(qg_tasks, key=len)) - 9
+                qg_header = (
+                    f"QuantumGraph contains {nQuanta} quanta for {len(qgraph.taskGraph)} tasks, "
+                    f"graph ID: '{qgraph.graphID}'"
+                )
+                qg_title = f"\033[31m Quanta  {'Task'.center(task_col_max_len)}\033[m"
```
timj, line 604: Please don't embed opaque color escape codes in print statements. No-one knows what they mean. I think we have a python package in rubin-env that does let you declare them by name, although I worry that using color here is a bit risky given you don't know the color of someone's terminal. Why do we need color?

leeskelvin: Colors were chosen to match the outputs of, e.g., `butler query-collections`. Is that not desirable? If not, I can revert to no color overrides. How/where is this handled by butler command line queries?

timj: In butler scripts we create an astropy Table and print it. It might be better for you to do that so you get more flexibility in the output rather than trying to do the padding yourself.

leeskelvin: I was hoping to avoid making multiple calls to the logger for this info block. That would ensure that the preceeding blurb (`INFO 2022-10-10T14:49:27.472-07:00 lsst.ctrl.mpexec.cmdLineFwk ()(cmdLineFwk.py:606)`) only gets printed once.

When I switch to using an astropy table and adding it into the logger, e.g.: `_LOG.info([qg_header] + [qg_table])`, I get output that looks like this:
```
INFO 2022-10-10T14:49:27.472-07:00 lsst.ctrl.mpexec.cmdLineFwk ()(cmdLineFwk.py:606) - ["QuantumGraph contains 10 quanta
 for 5 tasks, graph ID: '1665438567.4693098-3251389'", <Table length=5>
Quanta          Tasks         
int64           str23         
------ -----------------------
     2                     isr
     2       characterizeImage
     2               calibrate
     2     writePreSourceTable
     2 transformPreSourceTable]
```
which doesn't seem optimal. Is there a way to get around this?

timj: Don't you want to use `str()` there? Inside a stringifying a list results in each element using `repr()`.

taranu: Is the purpose of the `_LOG.isEnabledFor` clause just to be able to use f-strings (which I thought we should avoid to keep log message evaluation deferred)? It seems like you could keep a single call like `_LOG.info("QG has %d quanta for %d tasks, graph ID: %r\n%s", nQuanta, len(qgraph.taskGraph), qgraph.graphID, str(get_qgraph_task_table()))`. Either way I suggest making a separate function to return the table.

timj: The purpose of the isEnabledFor is that I was worried this could be a non trivial calculation and we should only do it if we know the message will be issued

leeskelvin: Thanks for the feedback both. I've just pushed with these edits in place:
* Astropy Table is now used to collect and print the results on screen
* Reverted back to the former logger approach (avoiding use of f-strings)
* Made a separate convenience method, `_generateTaskTable`, which does the heavy lifting in generating these statistics.

In addition, over on the sibling `pipe_base` PR (https://github.com/lsst/pipe_base/pull/285), an extra method (`getNumberOfQuantaForTask`) has been added which simplifies and speeds up this kind of number generation exercise. In testing for RC2 step 1 (~220000 quanta), this additional `for` loop takes ~0.01 seconds when using the older `len(qnodes)` framework, and a staggeringly fast ~0.001 seconds when using the newer `getNumberOfQuantaForTask` method.

taranu (approved):

PR #210, opened 2022-10-26
DM-36721: Ensure QG task table is printed in its entirety

leeskelvin:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/ctrl/mpexec/cmdLineFwk.py
+++ b/python/lsst/ctrl/mpexec/cmdLineFwk.py
@@ -594,12 +594,13 @@ class CmdLineFwk:
         else:
             if _LOG.isEnabledFor(logging.INFO):
                 qg_task_table = self._generateTaskTable(qgraph)
+                qg_task_table_formatted = "\n".join(qg_task_table.pformat_all())
                 _LOG.info(
                     "QuantumGraph contains %d quanta for %d tasks, graph ID: %r\n%s",
                     nQuanta,
                     len(qgraph.taskGraph),
                     qgraph.graphID,
-                    str(qg_task_table),
+                    qg_task_table_formatted,
                 )
 
         if args.save_qgraph:


python/lsst/ctrl/mpexec/cmdLineFwk.py
```diff
@@ -594,12 +594,13 @@ def makeGraph(self, pipeline: Pipeline, args: SimpleNamespace) -> Optional[Quant
         else:
             if _LOG.isEnabledFor(logging.INFO):
                 qg_task_table = self._generateTaskTable(qgraph)
+                qg_task_table_formatted = "\n".join(qg_task_table.pformat_all())
```
timj, line 597: Are you sure there isn't an astropy Table printing option for print all?

leeskelvin: `pformat` is an astropy formatting option (https://docs.astropy.org/en/stable/table/access_table.html#pformat-method). Are you thinking of `pprint`? That method prints the table to stdout immediately.

python/lsst/ctrl/mpexec/cmdLineFwk.py
```diff
@@ -734,14 +735,16 @@ def runPipeline(
                             # Do not save fields that are not set.
                             out.write(report.json(exclude_none=True, indent=2))
 
-    def _generateTaskTable(self, qgraph: QuantumGraph) -> Table:
+    def _generateTaskTable(self, qgraph: QuantumGraph, task_max_len: int = 72) -> Table:
```
parejkoj, line 738: I'm uncomfortable with this `task_max_len` value for something that is sent to a log. This goes back to the question of whether this output gets piped to a file/written to json, or just scrolled past on a screen during a run. A preserved log should not be truncated, while something written to the screen should let the terminal handle any excessive width in whatever manner it sees fit.

What would happen if you didn't set this value? Astropy's table formatter shouldn't know how wide the screen is, since this is going to the log, so wouldn't it just print the whole thing and have the terminal do the wrap-around?

leeskelvin: Task name truncation functionality removed from this ticket.

parejkoj (approved):

PR #212, opened 2022-12-02
DM-31725: Update to deal with daf_butler query refactor.

TallJimbo:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/tests/test_simple_pipeline_executor.py
+++ b/tests/test_simple_pipeline_executor.py
@@ -187,7 +187,7 @@ class SimplePipelineExecutorTests(lsst.utils.tests.TestCase):
         )
 
         with self.assertLogs("lsst", level="INFO") as cm:
-            quanta = executor.run(register_dataset_types=True)
+            quanta = executor.run(register_dataset_types=True, save_versions=False)
         self._test_logs(cm.output, "dict", "dict", "dict", "dict")
 
         self.assertEqual(len(quanta), 2)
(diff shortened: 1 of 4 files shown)


tests/test_simple_pipeline_executor.py
```diff
@@ -214,20 +214,27 @@ def test_from_pipeline_intermediates_differ(self):
...
         self.assertEqual(self.butler.get("intermediate").to_dict(), {"zero": 0, "one": 1})
+
+        # TODO: this test fails because we've set things up to need to do a
+        # dict -> TaskMetadata conversion on write and a TaskMetadata -> dict
+        # conversion on read, but the JSON formatter natively returns a dict
+        # and it looks like the formatter logic sees that as the same as the
+        # desired type and thinks it's done.  The problem is that the
+        # TaskMetadata dict representation has some entries that the dict it
+        # converts to does not.
+        self.skipTest("Upstream bug in daf_butler or TaskMetadata.")
```
timj, line 237: It might be better to mark this as expectedFail rather than skipping it. At least then when I fix it in daf_butler the test will fail because it passes.

python/lsst/ctrl/mpexec/singleQuantumExecutor.py
```diff
@@ -485,6 +485,7 @@ def updatedQuantumInputs(self, quantum: Quantum, butler: Butler, taskDef: TaskDe
             newRefsForDatasetType = updatedInputs[key]
             for ref in refsForDatasetType:
                 if ref.id is None:
+                    print(f"DM-31725: searching for ref {ref}.")
```
timj, line 488: Debug print?

tests/test_simple_pipeline_executor.py
TallJimbo: Looks like unittest can only do expected failures of an entire `TestCase`, so I've dropped the skip and changed from `assertEqual` to `assertNotEqual`.  That will have the same effect of forcing a fix to this test when the behavior is fixed, even though it won't appear as nicely in the categorization of the tests.

PR #214, opened 2022-12-15
DM-33497: Implement pipetask execution with quantum-backed butler

andy-slac:
Adds new quantum executor class and two new `pipetask` commands to execute quantum graph using quantum-backed butler.

## Checklist

- [X] ran Jenkins


python/lsst/ctrl/mpexec/cmdLineFwk.py
```diff
@@ -797,3 +813,86 @@ def _importGraphFixup(self, args: SimpleNamespace) -> Optional[ExecutionGraphFix
                 raise ValueError("Graph fixup is not an instance of ExecutionGraphFixup class")
             return fixup
         return None
+
+    def preExecInitQBB(self, task_factory: TaskFactory, args: SimpleNamespace) -> None:
+
+        # Load quantum graph. We do not really need individual Quanta here,
+        # but we need datastore records for initInputs, and those are only
+        # available from Quanta, so load the whole thing.
```
TallJimbo, line 821: Probably not worth doing it now, but I think this is another argument for making PreExecInit a special Quantum in the future.

python/lsst/ctrl/mpexec/preExecInit.py
```diff
@@ -396,3 +415,217 @@ def savePackageVersions(self, graph: QuantumGraph) -> None:
                     self.butler.put(oldPackages, datasetType, dataId)
             else:
                 self.butler.put(packages, datasetType, dataId)
+
+
+class PreExecInitLimited:
+    """Initialization of registry for QuantumGraph execution.
+
+    This class works with LimitedButler and expects that all references in
+    QuantumGraph are resolved.
```
TallJimbo, line 424: (not about this line, just picking a spot to start a thread)

It's a shame we now need two versions of each of `PreExecInit`, `SingleQuantumExecutor`, and task factories.  I wonder if we could consolidate a lot of code by having single implementations that:
 - are constructed with a `LimitedButler` (which may or may not be a full one);
 - attempt to cast their butler to a full one if they are given unresolved refs, and use the butler's registry to resolve them up front - first with `findDataset` (always works for execution butler, also works for inputs with regular butler), then with `DatasetIdFactory.resolveRef` if that returns `None`;
 - proceed with using `LimitedButler` interfaces only using those resolved refs.

For `SingleQuantumExecutor` we could probably get away with transforming the `Quantum` up-front by resolving all dataset refs.  It's less clear that something similar would work for pre-exec-init and task factories right now, and that's yet another argument for special pre-exec-init quanta in the future.

Right now it seems like we're not taking advantage of the fact that `Butler` _is a_ `LimitedButler` at all.  I think it's likely that this proposal would require moving some of the butler-creation responsibilities out of these objects and into `CmdLineFwk` or `MpQuantumExecutor`, but I think that might still be an overall win.

python/lsst/ctrl/mpexec/singleQuantumExecutor.py
```diff
@@ -478,7 +476,15 @@ def updatedQuantumInputs(self, quantum: Quantum, taskDef: TaskDef) -> Quantum:
         for key, refsForDatasetType in quantum.inputs.items():
             newRefsForDatasetType = updatedInputs[key]
             for ref in refsForDatasetType:
-                if ref.id is None:
+
+                # Inputs may already be resolved even if they do not exist, but
+                # we have to re-resolve them because IDs are ignored on output.
+                # If they already exist in datastore do not re-resolve them
+                # as it may change dataset type.
```
TallJimbo, line 483: As of DM-37450 (merged today) `findDataset` will respect the storage class of a `DatasetType` that is given to it.

TallJimbo (approved):
I'm a bit concerned about code duplication here, but I'm also willing to put fixing that off to another ticket in order to get things into a working state first and then consider whether any planned deprecations or interface changes might help with that.

python/lsst/ctrl/mpexec/cmdLineFwk.py
andy-slac: 👍 I want to avoid changing persistent QGraph format on this branch, I feel like it's not quite trivial and deserves a dedicated ticket.

python/lsst/ctrl/mpexec/preExecInit.py
andy-slac: There is indeed a bunch of duplicate code between two executors and I'm not very happy about it. I can try to refactor things a bit and move duplicated parts into a common base class used by both of those. I would want to avoid doing `isinstance` to check Butler type, I think better OO approach is to hide those differences in subclasses (and we probably still have a mock Butler class in unit tests that is not `Butler` subclass). What worries me a bit is that if we want to use `LimitedButler` API for most things, then some of the `Butler` functionality would need to be re-implemented in ctrl_mpexec classes.

python/lsst/ctrl/mpexec/singleQuantumExecutor.py
andy-slac: Hmm, I think I want to keep the datastore existence check here, otherwise I'd have to provide `timespan` to `findDataset` method which I don't know where to get. I'll update the comment.

TallJimbo (approved):
Looks good.  I can see why you might like aspects of this version as much as the previous one, but I think it's better in one very important way: it encapsulates more of the branching for the two butler types, so when we do (as I very much hope) move everything to QBB or at least make all DatasetRefs in QGs resolved, it will happen (mostly) behind-the-scenes.

PR #215, opened 2023-01-03
DM-37450: Fix text with faulty expectations.

TallJimbo:
New butler code causes the storage class in the DatasetType passed to findDataset to be respected, which causes the storage class in the input connection to be respected.

Relies on lsst/daf_butler#766

## Checklist

- [x] ran Jenkins


timj (approved):
Looks okay. Seems as if there are a couple of mypy fixes needed from my additions to daf_butler before Christmas.

TallJimbo:
I was worried I'd have to wade back into the `types-click` problem to deal with those annotations, but happily they were all on our end (one commit here and one in daf_butler).

PR #218, opened 2023-01-26
DM-34420: Coverage option

cbanek:
## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/ctrl/mpexec/cmdLineFwk.py
+++ b/python/lsst/ctrl/mpexec/cmdLineFwk.py
@@ -31,6 +31,8 @@ import copy
 import datetime
 import getpass
 import logging
+import os
+import pathlib
 import shutil
 from collections.abc import Iterable, Sequence
 from types import SimpleNamespace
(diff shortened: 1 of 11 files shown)


python/lsst/ctrl/mpexec/cmdLineFwk.py
```diff
@@ -757,6 +765,11 @@ def runPipeline(
                             # Do not save fields that are not set.
                             out.write(report.json(exclude_none=True, indent=2))
 
+        if args.coverage:
+            cov.stop()
+            cov.html_report(directory="covhtml")
```
timj, line 770: A textual summary sent to stdout would be useful as well. It looks like `cov.report()` can do that.

timj (changes requested):
Looks good. Needs a little bit more though:

* There is a `runGraphQBB` method in cmdLineFwk that needs the same changes made to it along with the associated click `run_qbb` command.
* We need to have a test so add something to `test_cmdLineFwk.py`. There are a few tests there that call `runPipeline` so making one of them set the coverage flag and then checking that a coverage directory was created would be okay.
* `coverage` needs to be listed as a dependency in `pyproject.toml`.

Have you checked that the coverage report looks reasonable and only includes our code?

PR #220, opened 2023-02-03
[DM-34420] Turn on coverage option

cbanek:
## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/ctrl/mpexec/cli/cmd/commands.py
+++ b/python/lsst/ctrl/mpexec/cli/cmd/commands.py
@@ -126,6 +126,7 @@ def build(ctx: click.Context, **kwargs: Any) -> None:
 
 @click.command(cls=PipetaskCommand, epilog=epilog)
 @click.pass_context
+@ctrlMpExecOpts.coverage_option()
 @ctrlMpExecOpts.show_option()
 @ctrlMpExecOpts.pipeline_build_options()
 @ctrlMpExecOpts.qgraph_options()
@@ -136,6 +137,11 @@ def build(ctx: click.Context, **kwargs: Any) -> None:
 def qgraph(ctx: click.Context, **kwargs: Any) -> None:
     """Build and optionally save quantum graph."""
     kwargs = _collectActions(ctx, **kwargs)
+    coverage = kwargs.pop("coverage", False)
+    if coverage:
+        print("Coverage turned on!", file=sys.stderr)
+    else:
+        print("Coverage not on", file=sys.stderr)
     show = ShowInfo(kwargs.pop("show", []))
     pipeline = script.build(**kwargs, show=show)
     if show.handled and not show.unhandled:
@@ -155,6 +161,12 @@ def qgraph(ctx: click.Context, **kwargs: Any) -> None:
 def run(ctx: click.Context, **kwargs: Any) -> None:
     """Build and execute pipeline and quantum graph."""
     kwargs = _collectActions(ctx, **kwargs)
+    coverage = kwargs.pop("coverage", False)
+    if coverage:
+        print("Coverage turned on!", file=sys.stderr)
+    else:
+        print("Coverage not on", file=sys.stderr)
+
     show = ShowInfo(kwargs.pop("show", []))
     pipeline = script.build(**kwargs, show=show)
     if show.handled and not show.unhandled:
@@ -236,6 +248,7 @@ def pre_exec_init_qbb(repo: str, qgraph: str, **kwargs: Any) -> None:
 @processes_option()
 @ctrlMpExecOpts.pdb_option()
 @ctrlMpExecOpts.profile_option()
+@ctrlMpExecOpts.coverage_option()
 @ctrlMpExecOpts.debug_option()
 @ctrlMpExecOpts.start_method_option()
 @ctrlMpExecOpts.timeout_option()
--- a/python/lsst/ctrl/mpexec/cli/cmd/commands.py
+++ b/python/lsst/ctrl/mpexec/cli/cmd/commands.py
@@ -19,13 +19,13 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-import os
-import pathlib
 import sys
+import tempfile
 from functools import partial
-from typing import Any, Tuple
+from typing import Any
 
 import click
+import coverage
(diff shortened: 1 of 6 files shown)


python/lsst/ctrl/mpexec/cli/cmd/commands.py
```diff
@@ -124,8 +126,33 @@ def build(ctx: click.Context, **kwargs: Any) -> None:
     _unhandledShow(show, "build")
 
 
+def start_coverage(coverage_packages: Tuple) -> Any:
```
timj, line 129: Can we annotate this to return a `coverage.Coverage` object? I think the `coverage_packages` type annotation can be `tuple[str, ...]`

Since this is an internal function maybe call it `_start_coverage` and `_stop_coverage` (if they are public we should document them).

python/lsst/ctrl/mpexec/cli/cmd/commands.py
```diff
@@ -124,8 +126,33 @@ def build(ctx: click.Context, **kwargs: Any) -> None:
...
+    print("Coverage turned ON!", file=sys.stderr)
```
timj, line 130: Maybe issue a single message combining the coverage package list. Also please remove the `!`. So an output along the lines of:

```
Coverage enabled of packages ...
```
or
```
Coverage enabled.
```

Since this is in the click layer I think we might want to use `click.echo()` rather than `print()` ?

python/lsst/ctrl/mpexec/cli/cmd/commands.py
```diff
@@ -124,8 +126,33 @@ def build(ctx: click.Context, **kwargs: Any) -> None:
...
+    if coverage_packages:
+        print("Coverage limited to: " + ",".join(coverage_packages), file=sys.stderr)
+
+    import coverage
+
+    coveragerc = os.path.join(pathlib.Path(__file__).parent.resolve(), "coveragerc")
+    cov = coverage.Coverage(
+        branch=True, concurrency="multiprocessing", config_file=coveragerc, source_pkgs=coverage_packages
```
timj, line 138: can't we say `config_file=False` here and then not need the coveragerc file at all? It looks like we are overriding the content anyhow.

python/lsst/ctrl/mpexec/cli/cmd/commands.py
```diff
@@ -19,9 +19,11 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import os
+import pathlib
 import sys
 from functools import partial
-from typing import Any
+from typing import Any, Tuple
```
timj, line 26: We can use python 3.10 in this package so using `tuple` is fine and no need to import from `typing`.

python/lsst/ctrl/mpexec/cli/cmd/commands.py
```diff
@@ -124,8 +126,33 @@ def build(ctx: click.Context, **kwargs: Any) -> None:
...
+    if coverage_packages:
+        print("Coverage limited to: " + ",".join(coverage_packages), file=sys.stderr)
+
+    import coverage
+
+    coveragerc = os.path.join(pathlib.Path(__file__).parent.resolve(), "coveragerc")
+    cov = coverage.Coverage(
+        branch=True, concurrency="multiprocessing", config_file=coveragerc, source_pkgs=coverage_packages
+    )
+    cov.load()
```
timj, line 140: It's this line that is loading previous coverage data. You need to think about what that means. It might be a useful feature if we understood how it worked.

python/lsst/ctrl/mpexec/cli/cmd/commands.py
```diff
@@ -124,8 +126,33 @@ def build(ctx: click.Context, **kwargs: Any) -> None:
...
+    cov.start()
+    return cov
+
+
+def stop_coverage(cov: Any) -> None:
+    cov.stop()
+    cov.html_report(directory="covhtml")
```
timj, line 147: Maybe do this as:

```python
outdir = "./covhtml"
cov.html_report(directory=outdir)
cov.report()
click.echo(f"Coverage data written to {outdir}")
```
so we don't write the path in twice.

python/lsst/ctrl/mpexec/cli/cmd/commands.py
```diff
@@ -236,6 +281,8 @@ def pre_exec_init_qbb(repo: str, qgraph: str, **kwargs: Any) -> None:
 @processes_option()
 @ctrlMpExecOpts.pdb_option()
 @ctrlMpExecOpts.profile_option()
+@ctrlMpExecOpts.coverage_option()
+@ctrlMpExecOpts.coverage_packages_option()
```
timj, line 285: The options have been added here but the code to turn on coverage is missing from the function and needs to be added.

timj (approved):
Looks great. There are some changes that need to be made to clean things up a little but I will mark it approved.

python/lsst/ctrl/mpexec/cli/cmd/commands.py
cbanek: Thanks for catching that, that's from the coverage docs that has that in there I think.  I'd rather just not have it load previous results as that is super confusing, at least it was for me.

python/lsst/ctrl/mpexec/cli/cmd/commands.py
cbanek: We need the coveragerc file because of the multiprocessing concurrency, otherwise it gives this error:

coverage.exceptions.ConfigError: multiprocessing requires a configuration file

python/lsst/ctrl/mpexec/cli/cmd/commands.py
timj: Okay. You might have to add it to the extra files included when the PyPI package is built since it's not a .py file so won't be found by default (like the py.typed file isn't). See the pyproject.toml file.

cbanek: I deleted the file and just am programmatically generating it from the code now since you said we may be overriding it, so I shouldn't need to add it anywhere else, it's just a string in there

cbanek: Lines 130-137

cbanek:
I've rebased and am ready to commit this, please shout if you don't want me to.

PR #224, opened 2023-02-28
DM-38165: Suppress traceback for a crashing task in unit test

andy-slac:
Apparently `faulthandler` is now enabled by default (which is a good thing in general) and it prints a long confusing traceback for the unit test that uses signal to kill a task. Disabling `faulthandler` in that task is sufficient to suppress traceback.

## Checklist

- [ ] ran Jenkins

### DIFF
--- a/tests/test_executors.py
+++ b/tests/test_executors.py
@@ -22,6 +22,7 @@
 """Simple unit test for cmdLineFwk module.
 """
 
+import faulthandler
 import logging
 import signal
 import sys
@@ -186,6 +187,8 @@ class TaskMockCrash:
 
     def runQuantum(self):
         _LOG.debug("TaskMockCrash.runQuantum")
+        # Disable fault handler to suppress long scary traceback.
+        faulthandler.disable()
         signal.raise_signal(signal.SIGILL)
 
 


timj (approved):

tests/test_executors.py
```diff
@@ -186,6 +187,8 @@ class TaskMockCrash:
 
     def runQuantum(self):
         _LOG.debug("TaskMockCrash.runQuantum")
+        # Disable fault handler to suppress long scary traceback.
+        faulthandler.disable()
```
kfindeisen, line 191: I'm not familiar with `faulthandler`, but this looks like a global setting. Don't you need to re-enable it when the test is complete (not sure how to do this if the task object is responsible for the setting, though).

kfindeisen (approved):

tests/test_executors.py
andy-slac: This is indeed a global setting for the whole process, but in this case it's fine because the process that runs the task is killed immediately on the next line. This task is only ever used in a forked execution (otherwise signal would kill the whole test) so it does not affect `faulthandler` in the parent process.

kfindeisen: Could you add a comment to that effect? It's not at all obvious in context that the task must only be run in an isolated environment.

andy-slac: Sorry, I pushed "Merge" already. I'll try to remember adding that on some other ticket.

PR #227, opened 2023-03-03
DM-38210: Use butler.get() rather than deprecated getDirect()

timj:
Requires lsst/daf_butler#797

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/ctrl/mpexec/singleQuantumExecutor.py
+++ b/python/lsst/ctrl/mpexec/singleQuantumExecutor.py
@@ -582,7 +582,7 @@ class SingleQuantumExecutor(QuantumExecutor):
                     ref = ref.unresolved()
                 self.butler.put(metadata, ref)
             else:
-                limited_butler.putDirect(metadata, ref)
+                limited_butler.put(metadata, ref)
 
     def initGlobals(self, quantum: Quantum) -> None:
         """Initialize global state needed for task execution.
--- a/tests/test_cmdLineFwk.py
+++ b/tests/test_cmdLineFwk.py
@@ -23,7 +23,6 @@
 """
 
 import contextlib
-import copy
 import logging
 import os
 import pickle
@@ -739,7 +738,7 @@ class CmdLineFwkTestCaseWithButler(unittest.TestCase):
         self.assertEqual(len(qgraph), self.nQuanta)
 
         # deep copy is needed because quanta are updated in place
-        fwk.runPipeline(copy.deepcopy(qgraph), taskFactory, args)
+        fwk.runPipeline(qgraph, taskFactory, args)
         self.assertEqual(taskFactory.countExec, self.nQuanta)
 
         # need to refresh collections explicitly (or make new butler/registry)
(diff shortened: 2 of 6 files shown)


python/lsst/ctrl/mpexec/singleQuantumExecutor.py
```diff
@@ -582,7 +582,7 @@ def writeMetadata(
                     ref = ref.unresolved()
                 self.butler.put(metadata, ref)
```
timj, line 583: @TallJimbo this fails if we don't unresolve the ref because of this test code for replace run in test_cmdLineFwk:

```python
        # re-run with --replace-run (--inputs is ignored, as long as it hasn't
        # changed)
        args.replace_run = True
        args.output_run = "output/run2"
        fwk.runPipeline(copy.deepcopy(qgraph), taskFactory, args)
```
and you get a resolved ref for run `output/run1` being put into `output/run2` and it failing because it wants to put it into `output/run1`. 

@andy-slac  Interestingly we also get another failure in the simplest possible test:

```
        qgraph = fwk.makeGraph(self.pipeline, args)
        # run whole thing
        fwk.runPipeline(qgraph, taskFactory, args)
```

where the problem is:

```
'output/20230330T204714Z' != 'output/20230330T204715Z'
```
with error: `No collection with name 'output/20230330T204714Z' found.` which makes me wonder how we are generating two different versions of that timestamped output run.

timj: Of course that second test failure with the timestamp is random because it depends on timing. I can also get `testSimpleQGraphClobberOutputs` to fail with the same timestamp mismatch as `testSimpleQGraph`.

andy-slac: If we don't unresolve refs then in `runPipeline` we should probably take output run from QuantumGraph. But I'm not sure how this will interact with all other options.

TallJimbo (approved):
I expected to see some changes that would prevent the `--replace-run` problem from happening in real usage (not just tests).  Did I just misunderstand whether that was necessary?

If not, could you just have it raise if you pass `--replace-run` while providing an existing QG?

timj:
I deliberately left the behavior the same (ie unresolving the DatasetRef with full butler). This allows replace-run to work as before. If I change this code to raise if the ref run and butler run are different that's a change in behavior that I assumed we wanted to avoid until @andy-slac tackled the removal of unresolved dataset ref (and is wrapped up in my comment on the jira ticket as future work).

TallJimbo:
Ah, sounds good - so we don't need to change anything now, but we will in order to drop unresolved refs.

tests/test_cmdLineFwk.py
```diff
@@ -762,7 +761,8 @@ def testSimpleQGraphReplaceRun(self):
         # changed)
         args.replace_run = True
         args.output_run = "output/run2"
-        fwk.runPipeline(copy.deepcopy(qgraph), taskFactory, args)
+        qgraph = fwk.makeGraph(self.pipeline, args)
```
timj, line 764: I made this change to demonstrate the fix but I wonder if we are punting to fixing this properly with unresolved ref removal that maybe I should not include this change here. @andy-slac would you rather this test failed once you remove unresolved refs or are you happy that this is the right fix for replace-run testing anyhow and so should keep this fix?

andy-slac: I'm OK with this change. I guess once we have the ability to rewrite output run in a graph we will add a new unit test for that.

PR #231, opened 2023-04-12
DM-38669: Add workaround for Python threading bug

andy-slac:
The bug causes occasional issues with multi-process pipeline executions with pytest using concurrent running. This is a known issue (https://github.com/python/cpython/issues/102512) which will eventually be fixed in Python, at that time the patch should become no-op and can be removed.

## Checklist

- [X] ran Jenkins


timj (approved):
I don't really understand it but agree that is definitely the workaround that other people are using for that bug.

PR #232, opened 2023-04-14
DM-37703: Catch unresolved refs warnings and some workarounds

timj:
Relies on lsst/daf_butler#821

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/ctrl/mpexec/mock_task.py
+++ b/python/lsst/ctrl/mpexec/mock_task.py
@@ -85,14 +85,9 @@ class MockButlerQuantumContext(ButlerQuantumContext):
             mockDatasetTypeName = self.mockDatasetTypeName(datasetType.name)
 
         try:
+            # Try to use the mock DatasetType if it is defined.
             mockDatasetType = self.butler.registry.getDatasetType(mockDatasetTypeName)
-            # Is the ref needed here? Can those parameters be passed directly
-            # to the get() call or is this ref needed in the KeyError except
-            # catch?
-            with warnings.catch_warnings():
-                warnings.simplefilter("ignore", category=UnresolvedRefWarning)
-                ref = DatasetRef(mockDatasetType, ref.dataId)
-            data = self.butler.get(ref)
+            data = self.butler.get(mockDatasetType, ref.dataId)
         except KeyError:
             data = super()._get(ref)
             # If the input as an actual non-mock data then we want to replace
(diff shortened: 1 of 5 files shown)


python/lsst/ctrl/mpexec/mock_task.py
```diff
@@ -85,7 +86,12 @@ def _get(self, ref: Optional[Union[DeferredDatasetRef, DatasetRef]]) -> Any:
 
         try:
             mockDatasetType = self.butler.registry.getDatasetType(mockDatasetTypeName)
-            ref = DatasetRef(mockDatasetType, ref.dataId)
+            # Is the ref needed here? Can those parameters be passed directly
+            # to the get() call or is this ref needed in the KeyError except
+            # catch?
```
andy-slac, line 91: Hmm, I think it should work if dateset type and dataId are passed to butler.get(), but ref is needed in `except` clause below for `super()._get()`. Also `KeyError` should probably be `LookupError`. I have not tested mock tasks for some time, there is a chance that all of this is somewhat broken.

tests/test_cmdLineFwk.py
```diff
@@ -274,7 +275,12 @@ def _makeQGraph():
             taskName=_TASK_CLASS,
             inputs={
                 fakeDSType: [
-                    DatasetRef(fakeDSType, DataCoordinate.standardize({"A": 1, "B": 2}, universe=universe))
+                    DatasetRef(
+                        fakeDSType,
+                        DataCoordinate.standardize({"A": 1, "B": 2}, universe=universe),
+                        run="fake_run",
+                        id=uuid.uuid4(),
```
andy-slac, line 282: You can drop this, I guess?

andy-slac (approved):
Looks good.

python/lsst/ctrl/mpexec/mock_task.py
timj: Right, so you create a ref in the try block specifically so it will overwrite the main ref and then be used in the except clause. Doesn't that mean the except clause is using mockDatasetType even though the comment after the `_get(ref)` call is implying it was a real dataset.

timj: I finally understand. The look up failure happens in the mockDatasetType definition itself. The change of the value of the `ref` variable is irrelevant. I was assuming the error came from the `butler.get`. That means the `ref` in the try block is not needed at all.

timj:
Classes similar to `MockRefcatDataId` are defined in a couple of tests in analysis_tools and drp_tasks. In theory we could modify the version here so it just takes a region and then remove then change the other packages so they use the one from here. Not entirely sure it's worth it but I'm happy to make the change.

PR #234, opened 2023-04-28
DM-38779: Try to fix output run inconsistency

timj:
This seems to be a real problem that is present in `pipetask run` itself but which we mask because the current code unresolves refs on put. The problem is that:

* makeGraph reads the supplied args, determines a time-stamped output run, and uses that to build the graph.
* runPipeline takes the graph and the args but uses the args to make a time-stamped output run.
* If the second has ticked over the output run is now different than the one used for the graph.
* singleQuantumExecutor runs and when Butler.put happens it unresolves the refs in the graph so that the timestamped RUN in the ref gets cleared and instead the butler's default RUN is used (and new UUID is issued).

I was expecting the unresolving to go away in #233 but we decided to leave that in place for now.

Fixing the tests is relatively easy. I will see if I can do a similar fix for `pipetask run` itself.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/ctrl/mpexec/cli/script/run.py
+++ b/python/lsst/ctrl/mpexec/cli/script/run.py
@@ -190,4 +190,10 @@ def run(  # type: ignore
 
     f = CmdLineFwk()
     taskFactory = TaskFactory()
+
+    # If we have no output run specified, use the one from the graph rather
+    # than letting a new timestamped run be created.
+    if not args.output_run and qgraphObj.metadata and (output_run := qgraphObj.metadata.get("output_run")):
+        args.output_run = output_run
+
     f.runPipeline(qgraphObj, taskFactory, args)
--- a/doc/changes/DM-38779.misc.rst
+++ b/doc/changes/DM-38779.misc.rst
@@ -1,0 +1,5 @@
+* ``SingleQuantumExecutor`` has been modified such that it no longer unresolves ``DatasetRef`` when putting the non-``PipelineTask`` datasets (such as packages and configs).
+  This has been done so that the refs in the quantum graph are preserved when they are written to a normal Butler.
+* Fixed a race condition when ``pipetask run`` creates the graph with a timestamped output run and then executes it.
+  Previously the graph creation and run execution phases calculated their own timestamped output run and it would be possible for the execution output run to be one second later than the graph run.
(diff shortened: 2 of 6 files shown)


timj:
I think it's possible that execution butler is now a bit broken from a provenance perspective. The graph is now required to have unresolved refs but we create execution butler by inserting the output datasets from the dataIDs and not from the resolve refs in the graph.

This leads to UUID mismatches and the log ingestion code currently fails because it gets confused when the UUID of the registry entry doesn't match the UUID of the thing being put.

TallJimbo:
I was assuming we were using the QG's refs to make the execution butler.  That seems like the problem, but I have no idea how hard it is to fix.  Should we consider rolling back some DatasetRef-resolving work until we can get rid of execution butler?

timj:
I will see if I can fix it next week

python/lsst/ctrl/mpexec/cmdLineFwk.py
```diff
@@ -211,7 +211,11 @@ def __init__(self, registry: Registry, args: SimpleNamespace, writeable: bool):
                     raise ValueError("Cannot use --extend-run option with non-existing or empty output chain")
                 runName = self.output.chain[0]
             else:
-                runName = "{}/{}".format(self.output, Instrument.makeCollectionTimestamp())
+                import random
+
+                extra = str(random.randint(0, 9))
```
timj, line 216: This is a temporary hack to guarantee that the code hits the race condition where we roll over to the next second in the timestamp calculation.

python/lsst/ctrl/mpexec/cli/script/run.py
```diff
@@ -190,4 +190,10 @@ def run(  # type: ignore
 
     f = CmdLineFwk()
     taskFactory = TaskFactory()
+
+    # If we have no output run specified, use the one from the graph rather
+    # than letting a new timestamped run be created.
+    if not args.output_run and qgraphObj.metadata:
+        args.output_run = qgraphObj.metadata["output_run"]
```
andy-slac, line 197: Maybe check for situation when metadata is present but "output_run" key is not? I guess we need a new ticket to make parts of metadata non-optional.

andy-slac (approved):
Looks good

PR #236, opened 2023-05-14
DM-39122: Remove UnresolvedRefWarning filters

andy-slac:
Additionally PreExecInit is cleaned up to always expect resolved refs in quantum graph and use those refs to store initOutputs. `CmdLineFwk` now checks that output runs on command line and in graph are consistent.

## Checklist

- [X] ran Jenkins

### DIFF
--- a/python/lsst/ctrl/mpexec/cli/opt/options.py
+++ b/python/lsst/ctrl/mpexec/cli/opt/options.py
@@ -63,11 +63,11 @@ pdb_option = MWOptionDecorator(
 
 extend_run_option = MWOptionDecorator(
     "--extend-run",
-    help=unwrap(
-        """Instead of creating a new RUN collection, insert datasets
-                                                  into either the one given by --output-run (if provided) or
-                                                  the first child collection of --output (which must be of
-                                                  type RUN). This also enables --skip-existing option."""
+    help=(
+        "Instead of creating a new RUN collection, insert datasets into either the one given by "
+        "--output-run (if provided) or the first child collection of --output (which must be of type RUN). "
+        "This also enables --skip-existing option when building a graph. "
+        "When executing a graph this option skips quanta with all existing outputs."
     ),
     is_flag=True,
 )
@@ -344,12 +344,10 @@ skip_existing_option = MWOptionDecorator(
 
 clobber_outputs_option = MWOptionDecorator(
     "--clobber-outputs",
-    help=unwrap(
-        """Remove outputs from previous execution of the same
-                                                       quantum before new execution.  If --skip-existing
-                                                       is also passed, then only failed quanta will be
-                                                       clobbered. Requires the 'run' command's --extend-run
-                                                       flag to be set."""
+    help=(
+        "Remove outputs of failed quanta from the output run when they would block the execution of new "
+        "quanta with the same data ID (or assume that this will be done, if just building a QuantumGraph).  "
+        "Does nothing if --extend-run is not passed."
     ),
     is_flag=True,
 )
--- a/python/lsst/ctrl/mpexec/preExecInit.py
+++ b/python/lsst/ctrl/mpexec/preExecInit.py
@@ -35,7 +35,7 @@ from typing import TYPE_CHECKING, Any
 # -----------------------------
 #  Imports for other modules --
 # -----------------------------
-from lsst.daf.butler import DataCoordinate, DatasetRef, DatasetType
+from lsst.daf.butler import DatasetRef, DatasetType
 from lsst.daf.butler.registry import ConflictingDefinitionError
 from lsst.pipe.base import PipelineDatasetTypes
 from lsst.utils.packages import Packages
@@ -83,9 +83,10 @@ class PreExecInitBase(abc.ABC):
     depend on Butler type.
     """
 
-    def __init__(self, butler: LimitedButler, taskFactory: TaskFactory):
+    def __init__(self, butler: LimitedButler, taskFactory: TaskFactory, extendRun: bool):
         self.butler = butler
         self.taskFactory = taskFactory
+        self.extendRun = extendRun
 
     def initialize(
         self,
@@ -166,14 +167,15 @@ class PreExecInitBase(abc.ABC):
             new data.
         """
         _LOG.debug("Will save InitOutputs for all tasks")
-        for taskDef in graph.iterTaskGraph():
-            init_input_refs = self.find_init_input_refs(taskDef, graph)
+        for taskDef in self._task_iter(graph):
+            init_input_refs = graph.initInputRefs(taskDef) or []
             task = self.taskFactory.makeTask(taskDef, self.butler, init_input_refs)
             for name in taskDef.connections.initOutputs:
                 attribute = getattr(taskDef.connections, name)
-                obj_from_store, init_output_ref = self.find_init_output(taskDef, attribute.name, graph)
+                init_output_refs = graph.initOutputRefs(taskDef) or []
+                init_output_ref, obj_from_store = self._find_dataset(init_output_refs, attribute.name)
                 if init_output_ref is None:
-                    raise ValueError(f"Cannot find or make dataset reference for init output {name}")
+                    raise ValueError(f"Cannot find dataset reference for init output {name} in a graph")
                 init_output_var = getattr(task, name)
 
                 if obj_from_store is not None:
@@ -220,10 +222,16 @@ class PreExecInitBase(abc.ABC):
         _LOG.debug("Will save Configs for all tasks")
         # start transaction to rollback any changes on exceptions
         with self.transaction():
-            for taskDef in graph.iterTaskGraph():
-                config_name = taskDef.configDatasetName
+            for taskDef in self._task_iter(graph):
+                # Config dataset ref is stored in task init outputs, but it
+                # may be also be missing.
+                task_output_refs = graph.initOutputRefs(taskDef)
+                if task_output_refs is None:
+                    continue
 
-                old_config, dataset_ref = self.find_init_output(taskDef, taskDef.configDatasetName, graph)
+                config_ref, old_config = self._find_dataset(task_output_refs, taskDef.configDatasetName)
+                if config_ref is None:
+                    continue
 
                 if old_config is not None:
                     if not taskDef.config.compare(old_config, shortcut=False, output=logConfigMismatch):
@@ -232,9 +240,10 @@ class PreExecInitBase(abc.ABC):
                             "butler; tasks configurations must be consistent within the same run collection"
                         )
                 else:
-                    # butler will raise exception if dataset is already there
-                    _LOG.debug("Saving Config for task=%s dataset type=%s", taskDef.label, config_name)
-                    self.butler.put(taskDef.config, dataset_ref)
+                    _LOG.debug(
+                        "Saving Config for task=%s dataset type=%s", taskDef.label, taskDef.configDatasetName
+                    )
+                    self.butler.put(taskDef.config, config_ref)
 
     def savePackageVersions(self, graph: QuantumGraph) -> None:
         """Write versions of software packages to butler.
@@ -254,7 +263,14 @@ class PreExecInitBase(abc.ABC):
 
         # start transaction to rollback any changes on exceptions
         with self.transaction():
-            old_packages, dataset_ref = self.find_packages(graph)
+            # Packages dataset ref is stored in graph's global init outputs,
+            # but it may be also be missing.
+
+            packages_ref, old_packages = self._find_dataset(
+                graph.globalInitOutputRefs(), PipelineDatasetTypes.packagesDatasetName
+            )
+            if packages_ref is None:
+                return
 
             if old_packages is not None:
                 # Note that because we can only detect python modules that have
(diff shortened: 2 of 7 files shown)


python/lsst/ctrl/mpexec/preExecInit.py
```diff
@@ -254,7 +263,14 @@ def savePackageVersions(self, graph: QuantumGraph) -> None:
 
         # start transaction to rollback any changes on exceptions
         with self.transaction():
-            old_packages, dataset_ref = self.find_packages(graph)
+            # Packages dataset ref is stored in task init outputs, but it
+            # may be also be missing.
```
TallJimbo, line 267: I would have thought it was that only the packages dataset ref is stored in the `globalInitOutputs`, while the task `initOutputs` are stored elsewhere in the graph.  But maybe the QuantumGraph interface just doesn't use the nomenclature I'd expected.

python/lsst/ctrl/mpexec/singleQuantumExecutor.py
```diff
@@ -387,7 +343,14 @@ def findOutputs(
                 checkRefs: list[DatasetRef] = []
                 registryRefToQuantumRef: dict[DatasetRef, DatasetRef] = {}
                 for datasetRef in datasetRefs:
-                    ref = self._resolve_ref(datasetRef, collections)
+                    if self.butler is not None:
+                        # If running with full butler check registry.
+                        ref = self.butler.registry.findDataset(
+                            datasetRef.datasetType, datasetRef.dataId, collections=collections
+                        )
```
TallJimbo, line 350: I don't understand why we'd want to re-resolve here.  What could lead to there being a dataset with a different UUID/run in the full-butler case that is impossible in the QBB case?

TallJimbo (approved):

python/lsst/ctrl/mpexec/preExecInit.py
andy-slac: That was copy-paste from `saveConfigs`. You are right, packages ref is in `globalInitOutputs`, I'll update the comment.

python/lsst/ctrl/mpexec/singleQuantumExecutor.py
andy-slac: It's not re-resolving, but merely checking for existing refs. This is the logic that implements skip-existing-in checks for `pipetask run`. I'm actually not very happy to have this logic here, but we have that option enabled for `pipetask run`. Maybe we should disable that option together with `output-run` and only have them enabled for quantum graph generation?

TallJimbo: I had thought it was already only an option for `pipetask run` in the case where it was also making a QG, which I think it consistent with disabling that option when using an existing QG and removing this check

andy-slac: I'll add a check to CLI to issue a warning when skip-existing is set with existing QG. This code is also used to find datasets that will be clobbered if clobbering is requested. I think I can modify it to use more efficient `Datastore.mexist()` instead of querying registry with DataIds.

TallJimbo: Discussion on [DM-32940](https://jira.lsstcorp.org/browse/DM-32940) may be relevant for that `mexists` attempt.  I suspect `mexists` (like `datasetExists` may be defined in a not-quite-substitutable way between `LimitedButler` and `Butler`, and fixing that may be the final piece of the puzzle in being able to unify SingleQuantumExecutor and PreExecinit implementations for the different butler types.

timj: Don't both butlers use a real datastore though? So datastore.knows_these and datastore.mexists should work in both cases.

TallJimbo: Yes, I think that's right - using datastore directly would circumvent the need for a butler API that always does only that.

andy-slac:
@TallJimbo, could you check my last commit [18faa70](https://github.com/lsst/ctrl_mpexec/pull/236/commits/18faa708e8adb797ece9a36f56d35eeb5da3e994) which removes skip-existing-in from SingleQuantumExecutor (and replaces it with extend_run option). I re-started Jenkins to see if it breaks anything.

python/lsst/ctrl/mpexec/singleQuantumExecutor.py
```diff
@@ -402,11 +338,9 @@ def findOutputs(
                         f" collection={self.butler.run} existingRefs={existingRefs}"
                         f" missingRefs={missingRefs}"
                     )
-            elif existingRefs and self.clobberOutputs and not self.skipExistingIn:
-                # Clobber complete outputs if skipExistingIn is not specified.
-                _LOG.info("Removing complete outputs for task %s: %s", taskDef, existingRefs)
-                self.butler.pruneDatasets(existingRefs, disassociate=True, unstore=True, purge=True)
-                return False
+            elif existingRefs:
+                # Everyhting is here already.
```
TallJimbo, line 342: typo

TallJimbo (approved):
Looks good to me, and I think it's important that we get the behavior here consistent with the `pipetask` CLI docs (as this does).  But the Prompt Processing Framework also uses some of these lower-level Python interfaces (via `SeparablePipelineExecutor)` now, so I'd also like to hear from @kfindeisen on whether there are any changes here that would be problematic, particularly:

- There's a new mandatory `extend_run` argument to `SeparablePipelineExecutor` construction.
- `--clobber-outputs` without `--skip-existing` is no longer possible.

python/lsst/ctrl/mpexec/separablePipelineExecutor.py
```diff
@@ -84,9 +84,15 @@ class SeparablePipelineExecutor:
     butler : `lsst.daf.butler.Butler`
         A Butler whose ``collections`` and ``run`` attributes contain the input
         and output collections to use for processing.
+    extend_run : `bool`, optional
+        True if extending an existing run instead of creating a new empty run.
+        Quanta with complete outputs will be skipped. Partial outputs from
+        failed quanta will be removed if ``clobberOutputs`` is `True`,
+        otherwise an exception will be raised.
```
kfindeisen, line 91: I find this argument a bit confusing, and I think other users will too -- by design, `SeparablePipelineExecutor` does not do any collection management, and the concept of extending/not extending a run isn't well-defined in Prompt Processing. That could lead to a situation where the client doesn't know how this flag should be set.

I see you pass this flag to `SingleQuantumExecutor`. In the existing code, `PreExecInit` does its "extend" processing (i.e., tolerating existing datasets) if and only if `clobber_output` is set. Judging from the new docs added below, exactly the same reasoning applies to `SingleQuantumExecutor`.

python/lsst/ctrl/mpexec/separablePipelineExecutor.py
```diff
@@ -84,9 +84,15 @@ class SeparablePipelineExecutor:
...
     clobber_output : `bool`, optional
-        If set, the pipeline execution overwrites existing output files.
-        Otherwise, any conflict between existing and new outputs is an error.
+        If set, the pipeline execution overwrites existing output files from
+        previously failed quanta. Otherwise, any conflict between existing and
+        new outputs is an error.
```
kfindeisen, line 95: If one wanted to overwrite all outputs unconditionally, how would one do that now?

python/lsst/ctrl/mpexec/singleQuantumExecutor.py
```diff
@@ -78,17 +78,15 @@ class SingleQuantumExecutor(QuantumExecutor):
...
-        Accepts list of collections, if all Quantum outputs already exist in
-        the specified list of collections then that Quantum will not be rerun.
-        If `None` then butler output RUN collection is searched for existing
-        datasets. If empty list then there no check for existing outputs (which
-        could result in conflicts when datasets are stored).
+    extendRun : bool, optional
+        True if execution is extending an existing run instead of creating a
+        new empty run. Quanta with complete outputs will be skipped. Partial
+        outputs from failed quanta will be removed if ``clobberOutputs`` is
+        `True`, otherwise an exception will be raised.
```
kfindeisen, line 85: What is the behavior if `extendRun` is false?

kfindeisen (commented):
Thanks for pointing me to this PR, @TallJimbo! I do have some questions about the implications for Prompt Processing, since (for performance reasons) we use runs in a very different pattern from batch processing.

python/lsst/ctrl/mpexec/cli/opt/options.py
```diff
@@ -63,11 +63,11 @@
...
-    help=unwrap(
-        """Instead of creating a new RUN collection, insert datasets
-                                                  into either the one given by --output-run (if provided) or
-                                                  the first child collection of --output (which must be of
-                                                  type RUN). This also enables --skip-existing option."""
+    help=(
+        "Instead of creating a new RUN collection, insert datasets into either the one given by "
+        "--output-run (if provided) or the first child collection of --output (which must be of type RUN). "
+        "This also enables --skip-existing option when building a graph. "
+        "When executing a graph this option skips quanta with all existing outputs."
```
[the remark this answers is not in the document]
andy-slac: The check for all outputs is sort of accidental there. The first check is on metadata existence, if it does not exist we need to check and possibly prune partial outputs. If we find that all outputs actually exists then we also say that no re-run is needed. This of course will only happen if `saveMetadata` is False. But we'll be checking for partial outputs in any case.

python/lsst/ctrl/mpexec/separablePipelineExecutor.py
andy-slac: `PreExecInit` logic is a bit more complicated, but it does not care about `clobberOutputs` as it cannot clobber anything, it only knows about `extendRun`. When extending a run, we require that all in-memory initOutputs be identical or compatible with what is already stored in the output run. When `extendRun` is not set, `PreExecInit` verifies that none of the initOutputs is stored already. When `extendRun` is set, it tries to read existing initOutputs and compare them to in-memory data (at least for configs and packages). And it stores initOutpus that do not exist in either case.

`SingleQuantumExecutor` logic is simpler - if `extendRun` and `clobberOutputs` are both set then remove partial outputs. If `extendRun` is true then it skips quantum execution if full set of outputs exists (it does not remove full outputs). Previous version of the code could behave somewhat differently in case inconsistent options were given (e.g. `clobberOutputs` is True but `extendRun` is False).

From what you say it seems that you want to have an option to clobber full outputs from a successful quanta execution, I think this may require some collection management. @TallJimbo may provide more insight into that.

python/lsst/ctrl/mpexec/separablePipelineExecutor.py
andy-slac: In `pipetask` land we have a couple of options (--replace-run and --prune-replaced) to clean up run collection, but it only works on the whole collections, not individual quanta. It should be possible to write few lines of code to explicitly remove quanta outputs from Butler.

python/lsst/ctrl/mpexec/singleQuantumExecutor.py
andy-slac: If `extendRun` is not set then it will just try to write to the output run without any checks, which should fail if the same dataset is already there.

python/lsst/ctrl/mpexec/separablePipelineExecutor.py
TallJimbo: > an option to clobber full outputs from a successful quanta execution

I don't think it needs collection management.  But I think it does mean `SingleQuantumExecutor` would require `extendRun=True` to be paired with _either_ `skipExisting=True` _or_ `clobberOutputs=True` (or both), and that would mean walking back some of the changes you just made.

I do think it'd still be fine for `pipetask` to have `--extend-run` imply `--skip-existing` because I don't think we need the prompt-processing clobber-everything behavior there, and I want to minimize how much that mode is used (I'd like to find a way to not need it in prompt processing, too, but I have not dug into the problem to see know if that might be feasible).

python/lsst/ctrl/mpexec/singleQuantumExecutor.py
kfindeisen: Can you say that explicitly?

python/lsst/ctrl/mpexec/separablePipelineExecutor.py
kfindeisen: Clobbering full outputs would be useful (at least, it matches my expectation of what "clobber" means), but I don't think it's the essential issue here.

My concern is that `extend_run` does not make sense as part of the `SeparablePipelineExecutor` API, because it suggests that there's `pipetask`-style collection management going on. As I noted above, when delegating to lower-level components that _do_ have a concept of extending runs, I've used `clobber_outputs` as a proxy (on the grounds that that's what you'd pass if you were doing any sort of retry). From @andy-slac's description, it sounds like `skip_existing` would actually be a better proxy when calling `SingleQuantumExecutor`, and in hindsight perhaps that's what I should have used for deciding how `PreExecInit` handles prior outputs as well.

At the very least, I think the calls to `PreExecInit` and `SingleQuantumExecutor` need to use the same strategy -- either the explicit `extend_run` flag added here, or a (matching?) proxy that expresses when their "extend-run" logic would apply to externally managed runs.

TallJimbo: I think you've got it right that `extendRun` isn't a meaningful concept at this low level, but that might just be a naming issue - at this level that value is still very useful as a way to express whether or not there is any chance of a collision between to-be-written datasets and existing datasets.  When there isn't we know we don't need to search for existing datasets, that can be fairly expensive.  Saying that we're not extending the run is sufficient (but not necessary way) to guarantee no collisions.

I can think of four states in `SingleQuantumExecutor` and `PreExecInit` that we have already tried to capture with these kinds of arguments:

1. Higher-level code guarantees that there is no chance of a collision, which may be because it's a new RUN or because it knows the data IDs that are present are disjoint with the new ones (though the latter is impossible for PreExecInit, since its outputs have empty data IDs).  This is the `extendRun=False` case.

2. There may be collisions, and the to-be-written datasets are always what we want, so clobber the old ones.  I am a bit skeptical of this state because it implies the new ones would not be the same as the old ones, which should only be possible if the outputs are not deterministic given the configs and software versions (which should be fixed, since collections only happen within a RUN), and that'd be a problem.  This is the `clobberOutputs=True`, `skipExisting=False` case.

3. There may be collisions, we expect the to-be-written ones to be equivalent to the old ones, and we want to retry any quanta that were not successful before, so clobber outputs from incomplete/failed quanta only.  This is the `clobberOutputs=True`, `skipExisting=True` case.

4. There may be collisions, we expect the to-be-written ones to be equivalent to the old ones, and we want to preserve those failures (and hold all downstream quanta).  This is the old `skipExisting=True` with `clobberOutputs=False`.

I think we need to keep 1, 3, and 4, but writing this up has reminded me why I did feel comfortable dropping (2) before, at least if there isn't a need for it in Prompt Processing.

kfindeisen: From the list above, `skipExisting` and `clobberOutputs` can exist in any combination, but is `extendRun` actually an extra dimension of that space? For example, is the combination `extend=True, clobber=False, skip=False` valid?

andy-slac: Sorry for delay, some usual distractions from other projects. I also looked at how all these options are used in GraphBuilder and PreExecInit, and I'm ready to get back to SingleQuantumExecutor. I think I can make it work by just having `skipExisting` and `clobberOutputs` options (and no `extendRun`). The logic would be:
- if full outputs (metadata or other outputs) exist:
  - if `skipExisting` is True then quantum execution is skipped
  - otherwise if `clobberOutputs` is true they are deleted
  - otherwise an error generated
- else if partial inputs are found:
  - if `clobberOutputs` is true they are deleted
  - otherwise an error generated

One minor issue is that `skipExisting` is a "derivative" of the `skip_existing_in`. The latter is needed to pass it to `GraphBuilder`, so `SeparablePipelineExecutor` already defines it. In that sense, I think it would be better for `SingleQuantumExecutor` to accept `skip_existing_in` instead of `skip_existing` (and this is how it was before my last commit).

TallJimbo: I think that misses out on a potentially important optimization of not checking for existing outputs at all, but I don't remember what we're doing in that respect right now - if we don't have a branch for that possibility already, let's go with your plan and come back to this later.  If we do have a branch for that, I think we should find a way to keep it to avoid a performance regression.

andy-slac: OK, I have re-implemented everything which has undone many of the changes on my previous commit. `SingleQuantumExecutor` interface is now unchanged from `main`, this means no changes to `SeparablePipelineExecutor`. I squashed it into one commit [7e0685b](https://github.com/lsst/ctrl_mpexec/pull/236/commits/7e0685bd718993c1f030b3eb8310defab80a63c8) to reduce noise. @TallJimbo, could you have another look.

TallJimbo (approved):
Looks good.  Sorry I helped send you off on a wild-goose chase in the middle of this.

PR #238, opened 2023-05-19
DM-39294: Refresh directed graph colors

leeskelvin:
## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/ctrl/mpexec/dotTools.py
+++ b/python/lsst/ctrl/mpexec/dotTools.py
@@ -30,6 +30,7 @@ __all__ = ["graph2dot", "pipeline2dot"]
 # -------------------------------
 #  Imports of standard modules --
 # -------------------------------
+import html
 import io
 import re
 from collections.abc import Iterable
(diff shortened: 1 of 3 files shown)


timj:
Can you please add some motivation for this change to the commit message?

leeskelvin:
Added this commit message text:
> The current pipeline directed graph nodes use dark gray as their background color. It has been reported that it is difficult to read the black text on the dark gray background. This ticket updates the directed graph color scheme with an aim towards making node text easier to read. In the process of exploring what color schemes would be optimal to satisfy the aim of this ticket, it emerged that making use of the Rubin visual identity colors may be desirable. This will help to make LSST pipeline graphs more instantly recognizable as Rubin-associated products. Colors: https://rubin.canto.com/g/RubinVisualIdentity

python/lsst/ctrl/mpexec/dotTools.py
```diff
@@ -91,9 +108,10 @@ def _renderQuantumNode(
 
 def _renderDSTypeNode(name: str, dimensions: list[str], file: io.TextIOBase) -> None:
     """Render GV node for a dataset type"""
-    labels = [name]
+    pointSize = int(_ATTRIBS["defaultNode"].get("fontsize", 12)) + 4
```
andy-slac, line 111: Hard-coded values are usually not welcome, could you add "fontsize" attribute to `_ATTRIBS["task"]` instead?

python/lsst/ctrl/mpexec/dotTools.py
```diff
@@ -49,32 +49,49 @@
...
+    default_attribs = ", ".join([f'{key}="{val}"' for key, val in attribs.items()])
+    print(f"{type} [{default_attribs}];", file=file)
+
+
 def _renderNode(file: io.TextIOBase, nodeName: str, style: str, labels: list[str]) -> None:
     """Render GV node"""
-    label = r"\n".join(labels)
-    attrib_dict = dict(_STYLES[style], label=label)
-    attrib = ", ".join([f'{key}="{val}"' for key, val in attrib_dict.items()])
+    label = r"</TD></TR><TR><TD>".join(labels)
```
andy-slac, line 72: I'm not very familiar with how graphviz handles HTML, do you need to escape special HTML characters here (imagine someone adds an ampersand to task name)?

python/lsst/ctrl/mpexec/dotTools.py
leeskelvin: There already is a `fontsize` attribute being set in `defaultNode`, which will be applied to all nodes (including task nodes). The line you've highlighted here is a safety net that will try to get `fontsize` from `defaultNode`, but, if someone were to remove `fontsize` down the line for some reason, it will fall back to a default value of 12.

leeskelvin: If you're not happy with setting this default here, I can remove the `12` and simply allow the code to fail at some future date if/when someone were to remove the `fontsize` attribute from `defaultNode`. Perhaps failing then is safer in any case?

python/lsst/ctrl/mpexec/dotTools.py
leeskelvin: According to the GraphViz docs (https://graphviz.org/doc/info/shapes.html#html), yes, HTML strings *will* be processed like HTML input. Do we think this will be a problem for us?

python/lsst/ctrl/mpexec/dotTools.py
andy-slac: I was mostly thinking about `+ 4`, but now I see that `fontsize` to task attributes also changes size of all labels, while intent here is to change the size of task name only. Maybe a separate "task-name" key can be added to _ATTRIBS to hold that value? Default `12` does not do much here, indeed, may be worth dropping it or defining a separate module-level constant for it instead?

python/lsst/ctrl/mpexec/dotTools.py
andy-slac: I'm OK with HTML, but I think we need to do proper escaping of strings, just in case people start using fancy names. There is a `html.escape()` escape method that should take care of that.

python/lsst/ctrl/mpexec/dotTools.py
leeskelvin: How about I replace this logic entirely with an externally-set `_NODELABELPOINTSIZE` attribute, set to 18? It's not just tasks that are impacted, it also impacts dataset type label names too. The intent behind the `+4` was that, if the user did change `fontsize`, the labels would still be somewhat larger. However, I'm happy to forego that convenience if hardcoding it makes it easier to read/maintain?

PR #240, opened 2023-05-26
DM-38952: revive support for mocked pipeline execution

TallJimbo:
Related to lsst/pipe_base#339

## Checklist

- [x] ran Jenkins


python/lsst/ctrl/mpexec/cli/cmd/commands.py
```diff
@@ -126,44 +121,40 @@ def build(ctx: click.Context, **kwargs: Any) -> None:
...
+            cov.save()
+            if report:
+                outdir = "./covhtml"
+                cov.html_report(directory=outdir)
+                click.echo(f"Coverage report written to {outdir}.")
 
 
 @click.command(cls=PipetaskCommand, epilog=epilog)
 @click.pass_context
-@ctrlMpExecOpts.coverage_option()
```
timj, line 165: I'm a bit confused as to how coverage options can ever be enabled if we have removed them. Is it missing a:

```python
ctrlMpExecOpts.coverage_options()
```
?

python/lsst/ctrl/mpexec/cli/opt/options.py
```diff
@@ -38,6 +38,10 @@
 
 coverage_option = MWOptionDecorator("--coverage", help="Enable coverage output.", is_flag=True)
 
+coverage_report_option = MWOptionDecorator(
+    "--cov-report/--no-cov-report", help="Whether to produce an HTML coverage report.", default=True
```
timj, line 42:
```suggestion
    "--cov-report/--no-cov-report", help="If coverage is enabled, controls whether to produce an HTML coverage report.", default=True
```

python/lsst/ctrl/mpexec/cmdLineFwk.py
```diff
@@ -562,6 +562,11 @@ def makeGraph(self, pipeline: Pipeline, args: SimpleNamespace) -> Optional[Quant
             if args.show_qgraph_header:
                 print(QuantumGraph.readHeader(args.qgraph))
         else:
+            task_defs = list(pipeline.toExpandedPipeline())
+            if args.mock:
```
timj, line 566: Is it not possible to have a test that uses mocks here?

python/lsst/ctrl/mpexec/cli/cmd/commands.py
TallJimbo: They were added back in `optionGroups.py`, by adding them to the `qgraph_options` decorator called a few lines below.

FWIW, this file organization that effectively splits the definition of script signatures across 3-4 files drives me nuts, and it may be what I'm really bothered by when I say I dislike click (compared to argparse).  But here I'm just trying to follow the existing pattern.

python/lsst/ctrl/mpexec/cmdLineFwk.py
TallJimbo: Yes, but I'd really like to start considering ci_middleware test coverage as sufficiently close to in-package test coverage that we avoid the duplication (of both effort and test-execution time).  In-package test coverage is still slightly preferable when it's just as easy to write a complete test there, but if I can only write a minimal test in-package that's wholly redundant with an easier-to-write and much more thorough test in ci_middleware, I'd prefer to get in the habit of just doing the latter.

python/lsst/ctrl/mpexec/cli/cmd/commands.py
timj: We try to put the click definitions in one place and the implementation in another because it's easier to test the implementation that way.

python/lsst/ctrl/mpexec/cmdLineFwk.py
timj: That's fine but makes it trickier for middleware users who aren't using the rest of the pipelines code.

timj (approved):

python/lsst/ctrl/mpexec/cmdLineFwk.py
```diff
@@ -595,6 +596,15 @@ def makeGraph(self, pipeline: Pipeline, args: SimpleNamespace) -> Optional[Quant
                 from lsst.pipe.base.tests.mocks import mock_task_defs
 
                 task_defs = mock_task_defs(task_defs, unmocked_dataset_types=args.unmocked_dataset_types)
+            instrument_class: type | None = None
```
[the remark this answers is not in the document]
TallJimbo: It's a shape `pipeline.getInstrument` wasn't originally defined as just returning the instrument class; it's not like there's anything you can do with that string other than import it.  I'm planning an RFC to deprecate TaskDef shortly after DM-33027 lands; maybe I can include deprecating `Pipeline.getInstrument` while adding a `Pipeline.instrument` property that returns the class object directly.

For this ticket, what do you think about a `Pipeline.get_data_id` method as an alternative way to share even more logic here?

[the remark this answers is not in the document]
timj: I'm fine with `Pipeline.get_data_id`

PR #241, opened 2023-05-30
DM-39434: Move QBB factory to a separate class

andy-slac:
It was a nested method before and that interfered with pickle in multiprocessing on MacOS. Moved that to a separate callable class.

## Checklist

- [X] ran Jenkins

### DIFF
--- a/python/lsst/ctrl/mpexec/cmdLineFwk.py
+++ b/python/lsst/ctrl/mpexec/cmdLineFwk.py
@@ -32,7 +32,7 @@ import datetime
 import getpass
 import logging
 import shutil
-from collections.abc import Iterable, Sequence
+from collections.abc import Iterable, Mapping, Sequence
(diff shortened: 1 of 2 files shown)


python/lsst/ctrl/mpexec/cmdLineFwk.py
```diff
@@ -66,7 +66,15 @@
 from .singleQuantumExecutor import SingleQuantumExecutor
 
 if TYPE_CHECKING:
-    from lsst.daf.butler import DatastoreRecordData, LimitedButler, Quantum, Registry
+    from lsst.daf.butler import (
```
timj, line 69: We probably should add that coverage exception list from daf_butler pyproject.toml.

timj (approved):
Is it possible to pickle something in ctrl_mpexec to trigger this code?

PR #243, opened 2023-06-05
DM-39377: Assume that we always store task metadata

andy-slac:
## Checklist

- [X] ran Jenkins


timj (approved):

timj (approved):
Maybe add a misc change log entry noting that the saveMetadata config is now ignored.

PR #246, opened 2023-06-15
DM-39661: Add support for ExecutionResources

timj:
Requires lsst/pipe_base#344

Add `--cores-per-quantum` and `--memory-per-quantum` options.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/ctrl/mpexec/cli/opt/options.py
+++ b/python/lsst/ctrl/mpexec/cli/opt/options.py
@@ -383,6 +383,30 @@ enable_implicit_threading_option = MWOptionDecorator(
     is_flag=True,
 )
 
+cores_per_quantum_option = MWOptionDecorator(
+    "-n",
+    "--cores-per-quantum",
+    default=1,
+    help=unwrap(
+        """Number of cores available to each quantum when executing.
+        If '-j' is used each subprocess will be allowed to use this number of cores."""
+    ),
+    type=click.IntRange(min=1),
+)
+
+memory_per_quantum_option = MWOptionDecorator(
+    "--memory-per-quantum",
+    default="",
+    help=unwrap(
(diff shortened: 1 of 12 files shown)


python/lsst/ctrl/mpexec/cli/script/run.py
```diff
@@ -159,6 +161,12 @@ def run(  # type: ignore
         If `True`, do not disable implicit threading by third-party libraries.
         Implicit threading is always disabled during actual quantum execution
         if ``processes > 1``.
+    cores_per_quantum : `int`
+        Number of cores that can be used by each quantum. Defaults to 1.
```
andy-slac, line 165: There is no default for that argument.

python/lsst/ctrl/mpexec/cli/opt/options.py
```diff
@@ -383,6 +383,29 @@
...
+
+memory_per_quantum_option = MWOptionDecorator(
+    "--memory-per-quantum",
+    default="",
+    help=unwrap(
+        """Memory allocated for each quantum to use when executing.
+        If '-j' used each subprocess will be allowed to use this amount of memory.
+        Units are allowed and the default units for a plain integer are MB.
+        For example: '3GB', '3000MB' and '3000' would all result in the same
+        memory limit. Default is for no limit."""
```
andy-slac, line 405: People may read it as if `pipetask` actually limits memory use. Maybe mention that this is an advisory value passed to the tasks?

andy-slac (approved):
Looks good

python/lsst/ctrl/mpexec/cli/script/run.py
timj: Ah. Thanks. The defaulting is above and below but in this API everything is mandatory...

PR #247, opened 2023-06-20
DM-39672: make it possible to force failures in mock pipelines on the command-line

TallJimbo:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/ctrl/mpexec/cli/opt/options.py
+++ b/python/lsst/ctrl/mpexec/cli/opt/options.py
@@ -19,8 +19,13 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+from __future__ import annotations
+
+from collections.abc import Iterable, Mapping
+
 import click
 from lsst.daf.butler.cli.utils import MWOptionDecorator, MWPath, split_commas, unwrap
+from lsst.utils.doImport import doImportType
 
 butler_config_option = MWOptionDecorator(
     "-b", "--butler-config", help="Location of the gen3 butler/registry config file."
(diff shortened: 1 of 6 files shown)


python/lsst/ctrl/mpexec/cli/opt/options.py
```diff
@@ -440,6 +445,46 @@
...
+    callback=parse_mock_failure,
+    metavar="LABEL:EXCEPTION:WHERE",
+    default=None,
+    multiple=True,
+    help=unwrap(
+        """Specifications for tasks that should be configured to fail
+        when mocking execution.  This is a colon-separated 3-tuple,
+        where the first entry the task label, the second the
+        fully-qualified exception type (empty for ValueError, and the
+        third a string (usually quoted) of the form passed to --where,
```
andy-slac, line 482: "usually quoted" may be a bit confusing. I guess quoting is needed for shell only, not for the Python code that parses it?

andy-slac (approved):
Looks good

python/lsst/ctrl/mpexec/cli/opt/options.py
```diff
@@ -440,6 +445,46 @@
...
+) -> Mapping[str, tuple[str, type[Exception] | None]]:
+    """Parse the --mock-failure option values into the mapping accepted by
+    `lsst.pipe.base.tests.mocks.mock_task_defs`.
+    """
+    result: dict[str, tuple[str, type[Exception] | None]] = {}
+    if value is None:
+        return result
+    for entry in value:
+        try:
+            task_label, error_type_name, where = entry.split(":")
```
andy-slac, line 460: One more comment - what if `where` also contains colon (e.g. for slicing)?

python/lsst/ctrl/mpexec/cli/opt/options.py
TallJimbo: Yes, just for the shell.  Reworded to that effect.

python/lsst/ctrl/mpexec/cli/opt/options.py
TallJimbo: Good catch.  I've updated this to
```
task_label, error_type_name, *where_terms = entry.split(":")
where = ":".join(where_terms)
```
to guard against that.

andy-slac: `entry.split(":", 2)` is probably faster 🙂

TallJimbo: Neat, I didn't know about that argument.

PR #256, opened 2023-08-04
DM-39779: text-based graph visualization of pipelines

TallJimbo:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/ctrl/mpexec/cli/script/build.py
+++ b/python/lsst/ctrl/mpexec/cli/script/build.py
@@ -27,12 +27,21 @@
 
 from types import SimpleNamespace
 
+from lsst.daf.butler import Butler
+
 from ... import CmdLineFwk
 from ..utils import _PipelineAction
 
 
 def build(  # type: ignore
-    order_pipeline, pipeline, pipeline_actions, pipeline_dot, save_pipeline, show, **kwargs
+    order_pipeline,
+    pipeline,
+    pipeline_actions,
+    pipeline_dot,
+    save_pipeline,
+    show,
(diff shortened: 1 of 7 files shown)


python/lsst/ctrl/mpexec/cli/script/build.py
```diff
@@ -53,7 +62,14 @@ def build(  # type: ignore
...
         Descriptions of what to dump to stdout.
-    kwargs : `dict` [`str`, `str`]
+    butler_config : `str`, `dict`, or `lsst.daf.butler.Config`, optional
+        If `str`, `butler_config` is the path location of the gen3
+        butler/registry config file. If `dict`, `butler_config` is key value
+        pairs used to init or update the `lsst.daf.butler.Config` instance. If
+        `Config`, it is the object used to configure a Butler.
+        Only used to resolve pipeline graphs for --show pipeline-graph and
+        --show task-graph.
+    **kwargs
```
arunkannawadi, line 72: The current version still has the type as `dict` `[`str`, `str`]`, but the change does not. Is there a reason for that?

arunkannawadi (approved):

python/lsst/ctrl/mpexec/cli/script/build.py
TallJimbo: My understanding is that the right type for `**kwargs`  is the _value_ type only, since it's always a dict, and the keys are always strings.  And it wasn't actually true before that the values were all strings, and I think that's most appropriately handled in this case by leaving out the type from the docs entirely.

PR #261, opened 2023-08-17
DM-40392: Update unit test after fixing QuantumGraph.updateRun

andy-slac:
Depends on lsst/pipe_base#369 and lsst/daf_butler#882

## Checklist

- [x] ran Jenkins

### DIFF
--- a/tests/test_cmdLineFwk.py
+++ b/tests/test_cmdLineFwk.py
@@ -60,7 +60,7 @@ from lsst.daf.butler import (
     Registry,
 )
 from lsst.daf.butler.core.datasets.type import DatasetType
-from lsst.daf.butler.registry import ConflictingDefinitionError, RegistryConfig
+from lsst.daf.butler.registry import RegistryConfig
 from lsst.pipe.base import (
     Instrument,
     Pipeline,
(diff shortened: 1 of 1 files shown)


andy-slac:
@timj, I can't understand why do-not-merge-checker is not Required for this package while it is Required in pipe_base and daf_butler. `do_not_merge.yaml` look identical, is there other configuration somewhere which makes it  required?

timj:
I fixed it. There is a setting for branch protections. Each new action has to be explicitly listed as required.

tests/test_cmdLineFwk.py
```diff
@@ -590,15 +590,12 @@ def test_simple_qgraph_qbb(self):
...
-                n2 = transfer_from_graph(temp_graph.name, self.root, True, False, False)
-
-                # For reasons that have to be investigated the number of
-                # outputs the second time around is only 21.
-                self.assertEqual(n2, 31)
-                self.assertEqual(n1, n2)
+            n2 = transfer_from_graph(temp_graph.name, self.root, True, False, False)
+
+            # For reasons that have to be investigated the number of
+            # outputs the second time around is only 21.
```
timj, line 596: Given that the test passes for you I think this means that the 21 I was getting was caused by the bad graph problems and you can remove this comment. We can likely remove the first test and only test that `n1 == n2`.

timj (approved):

PR #272, opened 2023-11-01
DM-41321: Add try-except block to manifest checker so that failure to get butler logs does not make error reporting impossible

eigerx:
## Checklist

- [x] ran Jenkins


eigerx:
The rename of "logs" to "do_store_logs" doesn't work on the current stack, but without this the rename in pipe_base will break. The Jenkins run was successful (https://rubin-ci.slac.stanford.edu/blue/organizations/jenkins/stack-os-matrix/detail/stack-os-matrix/582/pipeline/). I suspect that the build here will succeed following a merge of the ticket of the same name in pipe_base.

timj:
The idea is to temporarily change the build branch in requirements.txt

kherner (approved):

PR #276, opened 2023-12-01
DM-41962: Use storage classes from QG in PreExecInit.

TallJimbo:
PipelineDatasetTypes does not preserve storage classes (that's why it's being deprecated).

## Checklist

- [x] ran Jenkins (failed due to something unrelated; `main` is broken in at least two ways right now, and this ticket fixes one of them)

### DIFF
--- a/tests/test_simple_pipeline_executor.py
+++ b/tests/test_simple_pipeline_executor.py
@@ -165,7 +165,12 @@ class SimplePipelineExecutorTests(lsst.utils.tests.TestCase):
         return executor
 
     def _test_logs(self, log_output, input_type_a, output_type_a, input_type_b, output_type_b):
-        """Check the expected input types received by tasks A and B"""
+        """Check the expected input types received by tasks A and B.
+
+        Note that these are the types as seen from the perspective of the task,
+        so they must be consistent with the task's connections, but may not be
+        consistent with the registry dataset types.
+        """
         all_logs = "\n".join(log_output)
         self.assertIn(f"lsst.a:Run method given data of type: {input_type_a}", all_logs)
         self.assertIn(f"lsst.b:Run method given data of type: {input_type_b}", all_logs)
@@ -191,12 +196,6 @@ class SimplePipelineExecutorTests(lsst.utils.tests.TestCase):
 
     def test_from_pipeline_intermediates_differ(self):
         """Run pipeline but intermediates definition in registry differs."""
-        executor = self._configure_pipeline(
-            NoDimensionsTestTask.ConfigClass,
-            NoDimensionsTestTask.ConfigClass,
-            storageClass_b="TaskMetadataLike",
-        )
-
         # Pre-define the "intermediate" storage class to be something that is
         # like a dict but is not a dict. This will fail unless storage
         # class conversion is supported in put and get.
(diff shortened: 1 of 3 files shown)


tests/test_simple_pipeline_executor.py
```diff
@@ -243,14 +237,21 @@ def test_from_pipeline_output_differ(self):
...
+            storageClass_a="TaskMetadataLike",
+        )
+
         with self.assertLogs("lsst", level="INFO") as cm:
             quanta = executor.run(register_dataset_types=True, save_versions=False)
-        # a has been told to return a TaskMetadata but will convert to dict.
+        # a has been told to return a TaskMetadata but this will convert to
+        # dict on read by b.
         # b returns a dict and that is converted to TaskMetadata on put.
         self._test_logs(cm.output, "dict", "lsst.pipe.base.TaskMetadata", "dict", "dict")
```
timj, line 251: This is saying:

The test in line 250 is saying that

* "a" was given a dict
* "a" returned a TaskMetadata
* "b" was given a dict
* "b" returned a dict

which implies that "intermediate" should be a TaskMetadata and output should be a dict?

tests/test_simple_pipeline_executor.py
```diff
@@ -243,14 +237,21 @@ def test_from_pipeline_output_differ(self):
             )
         )
 
+        executor = self._configure_pipeline(
```
timj, line 240: This has to move until after the dataset type has been registered? If that is the case shouldn't the other tests have this change?

tests/test_simple_pipeline_executor.py
```diff
@@ -243,14 +237,21 @@ def test_from_pipeline_output_differ(self):
...
-        # a has been told to return a TaskMetadata but will convert to dict.
+        # a has been told to return a TaskMetadata but this will convert to
+        # dict on read by b.
         # b returns a dict and that is converted to TaskMetadata on put.
         self._test_logs(cm.output, "dict", "lsst.pipe.base.TaskMetadata", "dict", "dict")
 
         self.assertEqual(len(quanta), 2)
-        self.assertEqual(self.butler.get("intermediate"), {"zero": 0, "one": 1})
+        self.assertEqual(self.butler.get("intermediate").to_dict(), {"zero": 0, "one": 1})
         self.assertEqual(self.butler.get("output").to_dict(), {"zero": 0, "one": 1, "two": 2})
```
timj, line 255: So "output" is a TaskMetadata? How is that consistent with "b returned a dict" test above and the default connections being "dict"?

timj (commented):
I have questions before I can review ☹️

tests/test_simple_pipeline_executor.py
TallJimbo: The log messages are all from the perspective of the task, and are consistent with the task's connections.  They don't show the conversions that happen at the end of `put` or the beginning of `get`.

tests/test_simple_pipeline_executor.py
TallJimbo: Good point. They probably don't matter, but they're at least unsafe.  Will do.

tests/test_simple_pipeline_executor.py
timj: Ok. It should likely use the more explicit TaskMetadata.from_dict() in the test to make it obvious.

TallJimbo:
I've added another commit that cleans up the test a bit; ready for another look.

timj (approved):

PR #277, opened 2023-12-06
DM-41606: Option to output pipetask report info to command line

eigerx:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/ctrl/mpexec/cli/script/report.py
+++ b/python/lsst/ctrl/mpexec/cli/script/report.py
@@ -28,9 +28,10 @@
 from lsst.daf.butler import Butler
 from lsst.pipe.base import QuantumGraph
 from lsst.pipe.base.execution_reports import QuantumGraphExecutionReport
+from astropy.table import Table
 
 
-def report(butler_config: str, qgraph_uri: str, output_yaml: str, logs: bool = True) -> None:
+def report(butler_config: str, qgraph_uri: str, output_yaml: str | None, logs: bool = True) -> None:
     """Write a yaml file summarizing the produced and missing expected datasets
     in a quantum graph.
 
@@ -54,4 +55,22 @@ def report(butler_config: str, qgraph_uri: str, output_yaml: str, logs: bool = T
     butler = Butler.from_config(butler_config, writeable=False)
     qgraph = QuantumGraph.loadUri(qgraph_uri)
     report = QuantumGraphExecutionReport.make_reports(butler, qgraph)
-    report.write_summary_yaml(butler, output_yaml, do_store_logs=logs)
+    if not output_yaml:
+        # this is the option to print to the command-line
+        summary_dict = report.to_summary_dict(butler, logs, human_readable=True)
+        quanta = Table()
+        datasets = Table()
+        dataset_table_rows = []
+        data_products = []
+        for task in summary_dict.keys():
+            for data_product in summary_dict[task]["outputs"]:
+                print(summary_dict[task]["outputs"][data_product])
+                dataset_table_rows.append(summary_dict[task]["outputs"][data_product])
+                data_products.append(data_product)
+        datasets = Table(dataset_table_rows)
+        datasets.add_column(data_products, index=0, name="DatasetType")
+        print(datasets)
+            # something about failed quanta outside the smaller loop
+            #print(Table(summary_dict[task]["outputs"]))
+    else:
+        report.write_summary_yaml(butler, output_yaml, do_store_logs=logs)
--- a/python/lsst/ctrl/mpexec/cli/script/report.py
+++ b/python/lsst/ctrl/mpexec/cli/script/report.py
@@ -24,53 +24,86 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
+import pprint
 
+import yaml
+from astropy.table import Table
 from lsst.daf.butler import Butler
 from lsst.pipe.base import QuantumGraph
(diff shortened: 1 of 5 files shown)


eigerx:
It looks like the mypy failure is due to ctrl_mpexec not knowing about the added option (human-readable) in the corresponding branch of pipe_base. This passes on Jenkins: https://rubin-ci.slac.stanford.edu/blue/organizations/jenkins/stack-os-matrix/detail/stack-os-matrix/739/pipeline

timj:
@eigerx fix mypy by temporarily editing requirements.txt to use the ticket branch rather than main.

requirements.txt
```diff
@@ -6,6 +6,8 @@ networkx
 git+https://github.com/lsst/resources@main#egg=lsst-resources
 git+https://github.com/lsst/daf_butler@main#egg=lsst-daf-butler
 git+https://github.com/lsst/utils@main#egg=lsst-utils
-git+https://github.com/lsst/pipe_base@main#egg=lsst-pipe-base
+git+https://github.com/lsst/pipe_base@tickets/DM-41606#egg=lsst-pipe-base
 git+https://github.com/lsst/pex_config@main#egg=lsst-pex-config
 sqlalchemy
+types-psutil
+types-psycopg2
```
timj, line 13: FYI types go in a file called `types.txt` and not in `requirements.txt`. The mypy action looks in a special place because they aren't real requirements.

I'm a bit confused about the psutil one since psutil is only used by test code and so shouldn't get involved in mypy (it also is in the wrong place in pyproject.toml -- looks like it's been in the wrong place since the beginning so not related to this PR). Also, psycopg2 is not used anywhere in ctrl_mpexec so maybe I need

eigerx: AH, ok. I will move these edits to types.txt. I googled the mypy error and the first suggestion I tried was `mypy --install-types` and those were what it installed. They didn't end up resulting in a "pass" from mypy here, though. The other suggested line, `python3 -m pip install types-PyYAML`, did work, and that was what installed `types-PyYAML`. So, I'll do a hard reset and move that line to types.txt. Thank you for the help!!

eigerx: @timj it looks like that worked! Thanks for the linting help :)

requirements.txt
```diff
@@ -6,6 +6,6 @@ networkx
 git+https://github.com/lsst/resources@main#egg=lsst-resources
 git+https://github.com/lsst/daf_butler@main#egg=lsst-daf-butler
 git+https://github.com/lsst/utils@main#egg=lsst-utils
-git+https://github.com/lsst/pipe_base@main#egg=lsst-pipe-base
+git+https://github.com/lsst/pipe_base@tickets/DM-41606#egg=lsst-pipe-base
```
kherner, line 9: Also here don't forget to change it back when merging.

python/lsst/ctrl/mpexec/cli/script/report.py
```diff
@@ -24,34 +24,86 @@
...
-    output_yaml : `str`
-        The name to be used for the summary yaml file.
-    logs : `bool`
-        Get butler log datasets for extra information.
-
-    See Also
-    --------
-    lsst.pipe.base.QuantumGraphExecutionReport.make_reports : Making reports.
-    lsst.pipe.base.QuantumGraphExecutionReport.write_summary_yaml : Summaries.
+        butler_config : `str`
```
kherner, line 47: The indenting looks a little funny here relative to the old doc string, like it's an extra 4 spaces over.

python/lsst/ctrl/mpexec/cli/script/report.py
```diff
@@ -24,34 +24,86 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
+import pprint
 
+import yaml
```
kherner, line 29: Switch the blank line order here. See [PEP8](https://peps.python.org/pep-0008/#imports) for import format.

python/lsst/ctrl/mpexec/cli/script/report.py
```diff
@@ -24,34 +24,86 @@
...
+from astropy.table import Table
 from lsst.daf.butler import Butler
```
kherner, line 31: Probably should be a blank line between the astropy and Butler imports too, if you're being a format stickler.

kherner (changes requested):
The new code itself generally looks OK; I would guess it wouldn't be that hard to figure out the testing complaints with a few additions (i.e. a lot of it could probably be fixed with one or two calls). We can talk about that next week though.

requirements.txt
timj: There is a GitHub action that prevents merging unless this has been removed.

python/lsst/ctrl/mpexec/cli/script/report.py
timj: We use isort and black for formatting so I think this is right. `pprint` is part of core python and `yaml` is not so a line separates them.

python/lsst/ctrl/mpexec/cli/script/report.py
timj: isort won't allow that.

python/lsst/ctrl/mpexec/cli/script/report.py
timj: Agreed. I'm kind of surprised that the numpydoc validator doesn't spot this -- maybe I've missed some configuration setting for numpydoc. `velin` (which you have to install from pypi) does notice and fix it.

kherner (approved):
The changes to increase the test coverage look good.

PR #279, opened 2023-12-13
DM-42116: Add numpydoc validation

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/ctrl/mpexec/util.py
+++ b/python/lsst/ctrl/mpexec/util.py
@@ -28,80 +28,12 @@
 """Few utility methods used by the rest of a package.
 """
 
-__all__ = ["profile", "printTable", "filterTasks", "subTaskIter"]
+__all__ = ["printTable", "filterTasks", "subTaskIter"]
 
-# -------------------------------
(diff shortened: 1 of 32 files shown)


python/lsst/ctrl/mpexec/util.py
```diff
@@ -28,90 +28,23 @@
...
-
-# ------------------------
-#  Exported definitions --
-# ------------------------
-
-
-@contextlib.contextmanager
-def profile(filename: str, log: logging.Logger | None = None) -> Iterator:
-    """Context manager for profiling with cProfile
-
```
dhirving, line 58: Just want to check and confirm that this is fully private and safe to remove

dhirving (approved):

python/lsst/ctrl/mpexec/util.py
timj: It is a helper function for the `--profile` option and is identical to the version in `lsst.utils`. I removed the code duplication and the function (and `lsst.ctrl.mpexec.util`) was never meant to be public.

PR #283, opened 2024-02-01
DM-42751: Add num_proc parameter to SeparablePipelineExecutor.run_pipeline()

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/doc/changes/DM-42751.api.rst
+++ b/doc/changes/DM-42751.api.rst
@@ -1,0 +1,2 @@
+``SeparablePipelineExecutor.run_pipeline`` has been modified to take a ``num_proc`` parameter to specify how many subprocesses can be used to execute the pipeline.
+The default is now ``1`` (no spawning), which is a change from the previous behavior of using 80% of the available cores.
--- a/python/lsst/ctrl/mpexec/separablePipelineExecutor.py
+++ b/python/lsst/ctrl/mpexec/separablePipelineExecutor.py
@@ -36,8 +36,6 @@ __all__ = [
 import datetime
 import getpass
 import logging
-import math
-import multiprocessing
 from collections.abc import Iterable, Mapping
 from typing import Any, Protocol
 
@@ -241,6 +239,7 @@ class SeparablePipelineExecutor:
         graph: lsst.pipe.base.QuantumGraph,
         fail_fast: bool = False,
         graph_executor: QuantumGraphExecutor | None = None,
+        num_proc: int = 1,
     ) -> None:
         """Run a pipeline in the form of a prepared quantum graph.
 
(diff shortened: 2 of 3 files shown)


python/lsst/ctrl/mpexec/separablePipelineExecutor.py
```diff
@@ -241,6 +239,7 @@ def run_pipeline(
         graph: lsst.pipe.base.QuantumGraph,
         fail_fast: bool = False,
         graph_executor: QuantumGraphExecutor | None = None,
+        num_proc: int = 1,
```
kfindeisen, line 242: While I'm fine with changing the default to 1, I'm less fine about encoding that default in a parameter. The point of `SeparablePipelineExecutor` is to let the user plug in customized components as needed. I'm now wondering how `fail_fast` got in there; it doesn't seem to be something requested on #230.

python/lsst/ctrl/mpexec/separablePipelineExecutor.py
```diff
@@ -257,6 +256,10 @@ def run_pipeline(
         graph_executor : `lsst.ctrl.mpexec.QuantumGraphExecutor`, optional
             A custom graph executor. By default, a new instance of
             `lsst.ctrl.mpexec.MPGraphExecutor` is used.
+        num_proc : `int`, optional
+            The number of processes that can be used to run the pipeline. The
+            default value ensures that no subprocess is created. To determine
+            the number of available cores use `multiprocessing.cpu_count`.
```
kfindeisen, line 262: Please include the same "only with default executor" disclaimer as for `fail_fast`. The documentation for `fail_fast` itself should also be reworded to account for the fact that execution is no longer parallel by default.

kfindeisen: I'd recommend dropping the comment about `multiprocessing.cpu_count`; it's not relevant to this method. (Put another way, why bring it up it here, and not in all the other places where users specify number of processes?)

kfindeisen (approved):
I'm a bit concerned about the proliferation of executor-specific arguments in the `run_pipeline` parameter list, but I can't deny the precedent. I've requested some documentation changes to keep the focus on SPE's intended usage.

doc/changes/DM-42751.api.rst
kfindeisen, line 1: I'd recommend merging this commit with "Allow number of processes".

doc/changes/DM-42751.api.rst
```diff
@@ -0,0 +1,2 @@
+``SeparablePipelineExecutor.run_pipeline`` has been modified to take a ``num_proc`` parameter to specify how many subprocesses can be used to execute the pipeline.
+The default is now ``1`` (no spawning), which is a change from the previous behavior or using 80% of the available cores.
```
kfindeisen, line 2:
```suggestion
The default is now ``1`` (no spawning), which is a change from the previous behavior of using 80% of the available cores.
```

kfindeisen (approved):
Oops, looked only at the first commit.

PR #284, opened 2024-03-15
DM-43060: Let checkExistingOutputs work without a full butler

mfisherlevine:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/ctrl/mpexec/singleQuantumExecutor.py
+++ b/python/lsst/ctrl/mpexec/singleQuantumExecutor.py
@@ -290,6 +290,9 @@ class SingleQuantumExecutor(QuantumExecutor):
         If only partial outputs exist then they are removed if
         ``clobberOutputs`` is True, otherwise an exception is raised.
 
+        The ``LimitedButler`` is used for everything, and should be set to
+        ``self.butler`` if no separate ``LimitedButler`` is available.
+
         Parameters
         ----------
         quantum : `~lsst.daf.butler.Quantum`
@@ -312,10 +315,6 @@ class SingleQuantumExecutor(QuantumExecutor):
         RuntimeError
             Raised if some outputs exist and some not.
         """
-        if not self.butler:
-            # Skip/prune logic only works for full butler.
-            return False
-
         if self.skipExisting:
             _LOG.debug(
                 "Checking existence of metadata from previous execution of label=%s dataId=%s.",
@@ -333,7 +332,7 @@ class SingleQuantumExecutor(QuantumExecutor):
         _LOG.debug(
             "Looking for existing outputs in the way for label=%s dataId=%s.", taskDef.label, quantum.dataId
         )
-        ref_dict = self.butler.stored_many(chain.from_iterable(quantum.outputs.values()))
+        ref_dict = limited_butler.stored_many(chain.from_iterable(quantum.outputs.values()))
         existingRefs = [ref for ref, exists in ref_dict.items() if exists]
         missingRefs = [ref for ref, exists in ref_dict.items() if not exists]
         if existingRefs:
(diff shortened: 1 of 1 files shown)


python/lsst/ctrl/mpexec/singleQuantumExecutor.py
```diff
@@ -312,10 +312,6 @@ def checkExistingOutputs(self, quantum: Quantum, taskDef: TaskDef, limited_butle
         RuntimeError
             Raised if some outputs exist and some not.
         """
-        if not self.butler:
-            # Skip/prune logic only works for full butler.
-            return False
```
TallJimbo, line 317: @andy-slac, do you recall why we made this shortcut?  At first I thought it must be because `LimitedButler` didn't have `pruneDatasets`...but it does.  So this is either something we should have done a while ago (but maybe thought we didn't need) or there's something deeper.

This is all coming up because @mfisherlevine now has a `LimitedButler` that wraps a another `LimitedButler` (which is actually a full one in this use case) and does in-memory caching, and he's got a use case for that to be able to clobber.

TallJimbo (approved):

timj:
The docstring for this method needs some cleaning up now that limited butler is not just for queries. Should the docstring say how the limited butler relates to self.butler as well?

python/lsst/ctrl/mpexec/singleQuantumExecutor.py
```diff
@@ -343,26 +339,28 @@ def checkExistingOutputs(self, quantum: Quantum, taskDef: TaskDef, limited_butle
...
+                    run = self.butler.run if hasattr(self.butler, 'run') else 'run is n/a on a limited butler'
                     raise RuntimeError(
                         f"Complete outputs exists for a quantum {quantum} "
                         "and neither clobberOutputs nor skipExisting is set: "
-                        f"collection={self.butler.run} existingRefs={existingRefs}"
+                        f"collection={run} existingRefs={existingRefs}"
                     )
             else:
                 # Partial outputs from a failed quantum.
+                run = self.butler.run if hasattr(self.butler, 'run') else 'run is n/a on a limited butler'
```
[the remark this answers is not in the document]
mfisherlevine: Entirely up to you - which would you prefer?

mfisherlevine:
Thanks Tim. This was all done with Jim, and I don't think I actually know enough about this to write something correct/that will make you happy. Could you or Jim want to just suggest how you think it should read and I'll fix it up?

TallJimbo:
> Should the docstring say how the limited butler relates to self.butler as well?

I think the docstring for `checkExistingOutputs` can just say that the limited butler argument is now used for all parameters, and should be `self.butler` if there is no separate `LimitedButler`.

python/lsst/ctrl/mpexec/singleQuantumExecutor.py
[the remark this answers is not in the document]
mfisherlevine: OK, removed!

python/lsst/ctrl/mpexec/singleQuantumExecutor.py
mfisherlevine: This is all passing, but I'll wait to hear back from @andy-slac before merging anything.

andy-slac: I _think_ this was because back then we needed access to `butler.datastore` for implementation of this method and limited butler did not provide that property. We had extended limited butler interface since then, so it should be OK now to allow limited butler here.

run_err.yaml
andy-slac, line 1: Looks like some left-over from debugging? Should it be here?

andy-slac (approved):
Looks great, but there is probably one extra file that got into this commit unintentionally?

run_err.yaml
TallJimbo: I'm pretty sure that started getting created by tests that don't clean up after themselves in a different ticket.  @mfisherlevine, if you can spot what's making it and make it clean up, all the better, but just adding it to `.gitignore` instead of committing it is probably sufficient.

timj: It comes from here: https://github.com/lsst/ctrl_mpexec/blob/main/python/lsst/ctrl/mpexec/cli/script/report.py#L105-L106

mfisherlevine: Ack, the perils of routinely typing `git add -A` in repos. I'll just remove it from this ticket, but also leave it out of the gitignore, so that it _doesn't_ just get ignored. Then, when @eigerx fixes it on the other branch, it'll just go away cleanly without leaving something vestigial in the `.gitignore`.

eigerx: Sounds good, sorry all!

eigerx: https://rubinobs.atlassian.net/browse/DM-43363

PR #285, opened 2024-03-22
DM-43363: Remove extraneous files from pipetask report tests

eigerx:
## Checklist

- [x] ran Jenkins


timj:
I think the coverage dropping might be that show_errors is now never True? In the future consider having a test that does write the error file and then have you read the error file in to make sure it has the content you want.

eigerx:
Thanks for the review @timj ! I made a ticket for next sprint to include Pydantic models to validate pipetask report output.

PR #159, opened 2021-12-21
DM-33034: move GraphViz dot utilities to pipe_base

TallJimbo:
## Checklist

- [ ] ran Jenkins
- [ ] added a release note for user-visible changes to `doc/changes`

This PR is based on a temporary branch because its merge is blocked by DM-33027 (mostly because of changes in pipe_base, rather than here, but it's easier to keep things in sync this way).  I'll rebase once DM-33027 lands.

Nothing to see here but code removal; that code reappears in the [pipe_base PR for this ticket](https://github.com/lsst/pipe_base/pull/222).

PR #292, opened 2024-06-06
DM-44647: Move pipeline-dot build from cmdLineFwk to builder

leeskelvin:
This commit moves the pipeline-dot build logic from cmdLineFwk into cli/script/build.py.
As part of this commit, a switch to using the same back-end display args parser as --show pipeline-graph is also made.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/ctrl/mpexec/cmdLineFwk.py
+++ b/python/lsst/ctrl/mpexec/cmdLineFwk.py
@@ -75,7 +75,7 @@ from lsst.utils import doImportType
 from lsst.utils.logging import getLogger
 from lsst.utils.threads import disable_implicit_threading
 
-from .dotTools import graph2dot, pipeline2dot
+from .dotTools import graph2dot
 from .executionGraphFixup import ExecutionGraphFixup
 from .mpGraphExecutor import MPGraphExecutor
 from .preExecInit import PreExecInit, PreExecInitLimited
@@ -581,9 +581,6 @@ class CmdLineFwk:
 
         if args.save_pipeline:
             pipeline.write_to_uri(args.save_pipeline)
-
-        if args.pipeline_dot:
-            pipeline2dot(pipeline, args.pipeline_dot)
 
         return pipeline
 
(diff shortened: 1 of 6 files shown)


python/lsst/ctrl/mpexec/cmdLineFwk.py
```diff
@@ -582,9 +582,6 @@ def makePipeline(self, args: SimpleNamespace) -> Pipeline:
         if args.save_pipeline:
             pipeline.write_to_uri(args.save_pipeline)
 
-        if args.pipeline_dot:
-            pipeline2dot(pipeline, args.pipeline_dot)
```
timj, line 586: I am now thinking how this PR relates to #159 given that this is the only usage of pipeline2dot and that PR was removing it completely.

TallJimbo: I think this fully supersedes #159.  I'm also planning to RFC dropping support for QuantumGraph GraphViz outputs, since my understanding is that those are always too big for `dot` to render in practice anyway, and then we can get rid of `dotTools.py` entirely.

leeskelvin:
I've removed the `DO NOT MERGE` commit from this PR. Note that the `mypy` check now fails. After this ticket has been approved, I'll merge the `pipe_base` branch to `main` first, then re-run the checks on this PR to ensure that all checks are successful prior to merge.

arunkannawadi (approved):

requirements.txt
```diff
@@ -6,6 +6,7 @@ networkx
 lsst-resources @ git+https://github.com/lsst/resources@main
 lsst-daf-butler @ git+https://github.com/lsst/daf_butler@main
 lsst-utils @ git+https://github.com/lsst/utils@main
-lsst-pipe-base @ git+https://github.com/lsst/pipe_base@main
+lsst-pipe-base @ git+https://github.com/lsst/pipe_base@tickets/DM-44647
 lsst-pex-config @ git+https://github.com/lsst/pex_config@main
 sqlalchemy
+numpy >= 1.17, <2
```
timj, line 12: Hopefully this can be removed now.

leeskelvin: Thanks! Commit dropped.

PR #295, opened 2024-07-03
DM-45113: Re-enable skipping and clobbering with LimitedButler

mfisherlevine:
## Checklist

- [x] ran Jenkins


TallJimbo:
For other middleware devs: I suggested this in paircoding and plan to do a careful review of the neighboring code that _hasn't_ changed before signing off on the PR, but at first glance I don't see a reason why this isn't safe.

PR #297, opened 2024-07-24
DM-45340: Improve exception handling in quantum executors

andy-slac:
This should solve an issue with exceptions happening at unpredictable locations in a SingleQuantumExecutor class. MPGraphExecutor in a single-process mode should also behave reasonably for random exceptions. Multi-process mode has more complex internal state. If a client of the code catches an exception and tries to reuse the executor, the state may be inconsistent, depending on where the exception happens.

## Checklist

- [X] ran Jenkins

### DIFF
--- a/python/lsst/ctrl/mpexec/mpGraphExecutor.py
+++ b/python/lsst/ctrl/mpexec/mpGraphExecutor.py
@@ -167,15 +167,26 @@ class _Job:
             CliLog.replayConfigState(logConfigState)
 
         quantum = pickle.loads(quantum_pickle)
+        report: QuantumReport | None = None
         try:
-            quantumExecutor.execute(task_node, quantum)
+            _, report = quantumExecutor.execute(task_node, quantum)
+        except Exception as exc:
+            _LOG.debug("exception from task %s dataId %s: %s", task_node.label, quantum.dataId, exc)
+            report = QuantumReport.from_exception(
+                exception=exc,
+                dataId=quantum.dataId,
+                taskLabel=task_node.label,
+            )
+            raise
         finally:
-            # If sending fails we do not want this new exception to be exposed.
-            try:
-                report = quantumExecutor.getReport()
-                snd_conn.send(report)
-            except Exception:
-                pass
+            if report is not None:
+                # If sending fails we do not want this new exception to be
+                # exposed.
+                try:
+                    _LOG.debug("sending report for task %s dataId %s", task_node.label, quantum.dataId)
+                    snd_conn.send(report)
+                except Exception:
+                    pass
 
     def stop(self) -> None:
         """Stop the process."""
@@ -480,9 +491,18 @@ class MPGraphExecutor(QuantumGraphExecutor):
 
             _LOG.debug("Executing %s", qnode)
             try:
-                self.quantumExecutor.execute(task_node, qnode.quantum)
+                _, quantum_report = self.quantumExecutor.execute(task_node, qnode.quantum)
+                if quantum_report:
+                    report.quantaReports.append(quantum_report)
                 successCount += 1
             except Exception as exc:
+                quantum_report = QuantumReport.from_exception(
+                    exception=exc,
+                    dataId=qnode.quantum.dataId,
+                    taskLabel=task_node.label,
+                )
+                report.quantaReports.append(quantum_report)
+
                 if self.pdb and sys.stdin.isatty() and sys.stdout.isatty():
                     _LOG.error(
                         "Task <%s dataId=%s> failed; dropping into pdb.",
@@ -521,10 +541,6 @@ class MPGraphExecutor(QuantumGraphExecutor):
                 # collection cycle is run, which can happen at unpredictable
                 # times, run a collection loop here explicitly.
                 gc.collect()
-
-                quantum_report = self.quantumExecutor.getReport()
-                if quantum_report:
-                    report.quantaReports.append(quantum_report)
 
             _LOG.info(
                 "Executed %d quanta successfully, %d failed and %d remain out of total %d quanta.",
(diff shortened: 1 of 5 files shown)


python/lsst/ctrl/mpexec/quantumGraphExecutor.py
```diff
@@ -74,23 +80,6 @@ def execute(self, task_node: TaskNode | TaskDef, /, quantum: Quantum) -> Quantum
         """
         raise NotImplementedError()
 
-    def getReport(self) -> QuantumReport | None:
-        """Return execution report from last call to `execute`.
```
kfindeisen, line 78: This still seems like a useful API?

python/lsst/ctrl/mpexec/singleQuantumExecutor.py
```diff
@@ -164,26 +163,19 @@ def __init__(
...
-            result = self._execute(task_node, quantum)
-            self.report = QuantumReport(dataId=quantum.dataId, taskLabel=task_node.label)
-            return result
-        except Exception as exc:
-            self.report = QuantumReport.from_exception(
-                exception=exc,
-                dataId=quantum.dataId,
-                taskLabel=task_node.label,
-            )
-            raise
```
kfindeisen, line 186: Now I see the problem -- the `BaseException` is bypassing this block, but the code assumes `self.report` has been initialized one way or another. The same would happen with the block(s) moved to `MpGraphExecutor`.

Since catching `BaseException` is strongly discouraged (classes that are not `Exception` are those where premature catching would be directly harmful), would it make more sense to respec `getReport` as raising if execution was not attempted or was interrupted, and have the calling code catch `RuntimeError`?

python/lsst/ctrl/mpexec/mpGraphExecutor.py
```diff
@@ -168,11 +168,19 @@ def _executeJob(
...
-            quantumExecutor.execute(task_node, quantum)
+            _, report = quantumExecutor.execute(task_node, quantum)
+        except Exception as exc:
+            _LOG.debug("exception from task %s dataId %s: %s", task_node.label, quantum.dataId, exc)
+            report = QuantumReport.from_exception(
+                exception=exc,
+                dataId=quantum.dataId,
+                taskLabel=task_node.label,
+            )
+            raise
```
kfindeisen, line 179: This is exactly the same code as was originally in `SingleQuantumExecutor`, and will fail for the same reason.

tests/test_executors.py
kfindeisen, line 1: These changes would make a lot more sense as one or more standalone commits instead of being bundled with the reporting changes. As it is, they can't be separated now that you need to make changes.

kfindeisen (changes requested):
This PR doesn't actually fix the problem -- the new code in `MpGraphExecutor` is exactly as vulnerable to interrupts as the old code in `SingleQuantumExecutor` was (raising in the first task instead of a later one was a red herring). Having now seen the code for myself, I suspect catching the exception from `getReport` is the best solution.

I also recommend doing updates, code style, and maintenance on separate commits from the main work; it [usually makes sense to do them first](https://developer.lsst.io/work/flow.html#git-commit-organization-best-practices).

andy-slac:
@kfindeisen, I am certain that this indeed fixes the problem. Can you reproduce the crash that you saw before?

python/lsst/ctrl/mpexec/mpGraphExecutor.py
```diff
@@ -168,11 +168,19 @@ def _executeJob(
...
         finally:
             # If sending fails we do not want this new exception to be exposed.
             try:
-                report = quantumExecutor.getReport()
+                _LOG.debug("sending report for task %s dataId %s", task_node.label, quantum.dataId)
                 snd_conn.send(report)
             except Exception:
                 pass
```
kfindeisen, line 186: Wait... you mean you're _relying_ on this block to catch the `UnboundLocalError` if `report` is not defined? 😬

python/lsst/ctrl/mpexec/mpGraphExecutor.py
andy-slac: It will fail, but failure should be handled reasonably at this level in `report()` method.

python/lsst/ctrl/mpexec/singleQuantumExecutor.py
andy-slac: Problem with `getReport()` was that it relied on the state (`self.report`) and that state is hard to manage consistently in the presence of random exceptions. It is much easier to get rid of that state and return everything from `execute()`.

tests/test_executors.py
andy-slac: I have split this commit into two.

python/lsst/ctrl/mpexec/mpGraphExecutor.py
andy-slac: Fixed in a commit I just force-pushed.

python/lsst/ctrl/mpexec/mpGraphExecutor.py
andy-slac: Sorry, my previous reply went to a different comment, this is now fixed on latest commit.

tests/test_executors.py
```diff
@@ -71,47 +80,35 @@ class QuantumExecutorMock(QuantumExecutor):
...
             # in multiprocess mode use shared list
             manager = Manager()
             self.quanta = manager.list()
-        self.report = None
-        self._execute_called = False
 
-    def execute(self, task_node, /, quantum):
+    def execute(  # type: ignore[override]
+        self, task_node: TaskNodeMock, /, quantum: QuantumMock  # type: ignore[override]
+    ) -> tuple[QuantumMock, QuantumReport | None]:
```
kfindeisen, line 92: This commit still mixes typing and report changes. This is exactly why it's best to commit them separately in the first place.

Unfortunately, I don't see an easy way to disentangle the commits now -- both sets of changes need to modify `test_executors.py`.

kfindeisen (approved):
Looks much better, thanks.

tests/test_executors.py
andy-slac: True, and I do not want to spend much time on trying to perfect it.

PR #303, opened 2024-08-23
DM-44488: handle new pipe_base exception types in executors

TallJimbo:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/ctrl/mpexec/cli/script/run.py
+++ b/python/lsst/ctrl/mpexec/cli/script/run.py
@@ -68,6 +68,7 @@ def run(  # type: ignore
     cores_per_quantum: int,
     memory_per_quantum: str,
     rebase,
+    raise_on_partial_outputs: bool,
     **kwargs,
 ):
     """Implement the command line interface `pipetask run` subcommand.
@@ -179,6 +180,8 @@ def run(  # type: ignore
     rebase : `bool`
         If `True` then reset output collection chain if it is inconsistent with
         the ``inputs``.
+    raise_on_partial_outputs : `bool`
+        Consider partial outputs an error instead of a success.
     **kwargs : `dict` [`str`, `str`]
         Ignored; click commands may accept options for more than one script
         function and pass all the option kwargs to each of the script functions
@@ -222,6 +225,7 @@ def run(  # type: ignore
         cores_per_quantum=cores_per_quantum,
         memory_per_quantum=memory_per_quantum,
         rebase=rebase,
+        raise_on_partial_outputs=raise_on_partial_outputs,
     )
 
     f = CmdLineFwk()
--- a/python/lsst/ctrl/mpexec/cli/script/run_qbb.py
+++ b/python/lsst/ctrl/mpexec/cli/script/run_qbb.py
@@ -52,6 +52,7 @@ def run_qbb(
     enable_implicit_threading: bool,
     cores_per_quantum: int,
     memory_per_quantum: str,
+    raise_on_partial_outputs: bool,
 ) -> None:
     """Implement the command line interface ``pipetask run-qbb`` subcommand.
 
@@ -101,6 +102,8 @@ def run_qbb(
         Amount of memory that each quantum can be allowed to use. Empty string
         implies no limit. The string can be either a single integer (implying
         units of MB) or a combination of number and unit.
+    raise_on_partial_outputs : `bool`
+        Consider partial outputs an error instead of a success.
     """
     # Fork option still exists for compatibility but we use spawn instead.
     if start_method == "fork":
@@ -127,6 +130,7 @@ def run_qbb(
         enable_implicit_threading=enable_implicit_threading,
         cores_per_quantum=cores_per_quantum,
         memory_per_quantum=memory_per_quantum,
+        raise_on_partial_outputs=raise_on_partial_outputs,
     )
 
     f = CmdLineFwk()
--- a/python/lsst/ctrl/mpexec/separablePipelineExecutor.py
+++ b/python/lsst/ctrl/mpexec/separablePipelineExecutor.py
@@ -79,13 +79,18 @@ class SeparablePipelineExecutor:
     skip_existing_in : iterable [`str`], optional
         If not empty, the pipeline execution searches the listed collections
         for existing outputs, and skips any quanta that have run to completion
-        (or have no work to do). Otherwise, all tasks are attempted (subject
-        to ``clobber_output``).
+        (or have no work to do). Otherwise, all tasks are attempted (subject to
+        ``clobber_output``).
     task_factory : `lsst.pipe.base.TaskFactory`, optional
         A custom task factory for use in pre-execution and execution. By
         default, a new instance of `lsst.ctrl.mpexec.TaskFactory` is used.
     resources : `~lsst.pipe.base.ExecutionResources`
         The resources available to each quantum being executed.
+    raise_on_partial_outputs : `bool`, optional
+        If `True` raise exceptions chained by
+        `lsst.pipe.base.AnnotatedPartialOutputError` immediately, instead of
+        considering the partial result a success and continuing to run
+        downstream tasks.
     """
 
     def __init__(
@@ -95,6 +100,7 @@ class SeparablePipelineExecutor:
         skip_existing_in: Iterable[str] | None = None,
         task_factory: lsst.pipe.base.TaskFactory | None = None,
         resources: lsst.pipe.base.ExecutionResources | None = None,
+        raise_on_partial_outputs: bool = False,
     ):
         self._butler = Butler.from_config(butler=butler, collections=butler.collections, run=butler.run)
         if not self._butler.collections:
@@ -107,6 +113,7 @@ class SeparablePipelineExecutor:
 
         self._task_factory = task_factory if task_factory else TaskFactory()
         self.resources = resources
+        self.raise_on_partial_outputs = raise_on_partial_outputs
 
     def pre_execute_qgraph(
         self,
@@ -271,6 +278,7 @@ class SeparablePipelineExecutor:
                 skipExistingIn=self._skip_existing_in,
                 clobberOutputs=self._clobber_output,
                 resources=self.resources,
+                raise_on_partial_outputs=self.raise_on_partial_outputs,
             )
             graph_executor = MPGraphExecutor(
                 numProc=num_proc,
--- a/python/lsst/ctrl/mpexec/simple_pipeline_executor.py
+++ b/python/lsst/ctrl/mpexec/simple_pipeline_executor.py
@@ -56,6 +56,11 @@ class SimplePipelineExecutor:
         ``quantum_graph``.
     resources : `~lsst.pipe.base.ExecutionResources`
         The resources available to each quantum being executed.
+    raise_on_partial_outputs : `bool`, optional
+        If `True` raise exceptions chained by
+        `lsst.pipe.base.AnnotatedPartialOutputError` immediately, instead of
+        considering the partial result a success and continuing to run
+        downstream tasks.
 
     Notes
     -----
@@ -79,10 +84,12 @@ class SimplePipelineExecutor:
         quantum_graph: QuantumGraph,
         butler: Butler,
         resources: ExecutionResources | None = None,
+        raise_on_partial_outputs: bool = False,
     ):
         self.quantum_graph = quantum_graph
         self.butler = butler
         self.resources = resources
+        self.raise_on_partial_outputs = raise_on_partial_outputs
 
     @classmethod
     def prep_butler(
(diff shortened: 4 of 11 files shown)


python/lsst/ctrl/mpexec/cli/script/run.py
```diff
@@ -179,6 +180,8 @@ def run(  # type: ignore
     rebase : `bool`
         If `True` then reset output collection chain if it is inconsistent with
         the ``inputs``.
+    raise_on_partial_outputs : `bool`
+        Consider partial outputs and error instead of a success.
```
parejkoj, line 184: Is this what you meant?
```suggestion
        Consider partial outputs an error instead of a success.
```

python/lsst/ctrl/mpexec/cli/script/run_qbb.py
```diff
@@ -101,6 +102,8 @@ def run_qbb(
         Amount of memory that each quantum can be allowed to use. Empty string
         implies no limit. The string can be either a single integer (implying
         units of MB) or a combination of number and unit.
+    raise_on_partial_outputs : `bool`
+        Consider partial outputs and error instead of a success.
```
parejkoj, line 106:
```suggestion
        Consider partial outputs an error instead of a success.
```

python/lsst/ctrl/mpexec/separablePipelineExecutor.py
```diff
@@ -79,13 +79,17 @@ class SeparablePipelineExecutor:
...
+        ``clobber_output``).
     task_factory : `lsst.pipe.base.TaskFactory`, optional
         A custom task factory for use in pre-execution and execution. By
         default, a new instance of `lsst.ctrl.mpexec.TaskFactory` is used.
     resources : `~lsst.pipe.base.ExecutionResources`
         The resources available to each quantum being executed.
+    raise_on_partial_outputs : `bool`, optional
+        If `True` raise exceptions chained by
+        `lsst.pipe.base.AnnotatedPartialOutputError` immediately, instead of
+        consider the partial result a success and running downstream tasks.
```
parejkoj, line 92:
```suggestion
        considering the partial result a success and continuing to run downstream tasks.
```

python/lsst/ctrl/mpexec/simple_pipeline_executor.py
```diff
@@ -56,6 +56,10 @@ class SimplePipelineExecutor:
         ``quantum_graph``.
     resources : `~lsst.pipe.base.ExecutionResources`
         The resources available to each quantum being executed.
+    raise_on_partial_outputs : `bool`, optional
+        If `True` raise exceptions chained by
+        `lsst.pipe.base.AnnotatedPartialOutputError` immediately, instead of
+        consider the partial result a success and running downstream tasks.
```
parejkoj, line 62:
```suggestion
        considering the partial result a success and continuing to run downstream tasks.
```

python/lsst/ctrl/mpexec/simple_pipeline_executor.py
```diff
@@ -159,16 +166,28 @@ def from_pipeline_filename(
             one.
         resources : `~lsst.pipe.base.ExecutionResources`
             The resources available to each quantum being executed.
+        raise_on_partial_outputs : `bool`, optional
+            If `True` raise exceptions chained by
+            `lsst.pipe.base.AnnotatedPartialOutputError` immediately, instead
+            of consider the partial result a success and running downstream
+            tasks.
```
parejkoj, line 173:
```suggestion
            of considering the partial result a success and continuing to run downstream
            tasks.
```

python/lsst/ctrl/mpexec/simple_pipeline_executor.py
```diff
@@ -205,6 +225,11 @@ def from_task_class(
...
```
parejkoj, line 232:
```suggestion
            of considering the partial result a success and continuing to run downstream
            tasks.
```

python/lsst/ctrl/mpexec/simple_pipeline_executor.py
```diff
@@ -257,6 +288,11 @@ def from_pipeline(
...
```
parejkoj, line 295:
```suggestion
            of considering the partial result a success and continuing to run downstream
            tasks.
```

python/lsst/ctrl/mpexec/simple_pipeline_executor.py
```diff
@@ -300,6 +342,11 @@ def from_pipeline_graph(
             empty and not `None`.
         resources : `~lsst.pipe.base.ExecutionResources`
             The resources available to each quantum being executed.
+        raise_on_partial_outputs : `bool`, optional
+            If `True` raise exceptions chained by
+            `lsst.pipe.base.AnnotatedPartialOutputError` immediately, instead
+            of consider the partial result a success and running downstream
+            tasks.
```
parejkoj, line 349:
```suggestion
            of considering the partial result a success and continuing to run downstream
            tasks.
```

python/lsst/ctrl/mpexec/singleQuantumExecutor.py
```diff
@@ -119,6 +120,10 @@ class SingleQuantumExecutor(QuantumExecutor):
         not look for them.  This causes the ``skipExisting`` and
         ``clobberOutputs`` options to be ignored, but unlike just setting both
         of those to `False`, it also avoids all dataset existence checks.
+    raise_on_partial_outputs : `bool`, optional
+        If `True` raise exceptions chained by
+        `lsst.pipe.base.AnnotatedPartialOutputError` immediately, instead of
+        consider the partial result a success and running downstream tasks.
```
parejkoj, line 126:
```suggestion
        considering the partial result a success and continuing to run downstream tasks.
```

python/lsst/ctrl/mpexec/singleQuantumExecutor.py
```diff
@@ -476,8 +477,39 @@ def runQuantum(
...
+                str(err),
+            )
+        except AnnotatedPartialOutputsError as caught:
+            error: BaseException
+            if caught.__cause__ is None:
+                _LOG.warning(
+                    "Incorrect use of AnnotatedPartialOutputsError: no chained exception found.",
+                    task_node.label,
+                    quantum.dataId,
+                )
```
parejkoj, line 493: Oh, I like that. I think it should maybe be an `error`, though, not a `warning`, so it's more visible? That seems like a serious coding problem that should be loud.

python/lsst/ctrl/mpexec/singleQuantumExecutor.py
```diff
@@ -476,8 +477,39 @@ def runQuantum(
...
+                error = caught
+            else:
+                error = caught.__cause__
+            if self.raise_on_partial_outputs:
+                # This is a real edge case that required some experimentation:
```
parejkoj, line 498: I think I'd put a `NOTE:` or something at the start of this, as it's sort of a meta-comment.

python/lsst/ctrl/mpexec/singleQuantumExecutor.py
```diff
@@ -476,8 +477,39 @@ def runQuantum(
...
+                # without 'from None' below, this raise would produce a
+                # "while one exception was being handled, another was raised"
+                # traceback involving AnnotatedPartialOutputsError.  With
+                # the 'from None', we get just the error chained to it, as
+                # desired.
+                raise error from None
+            else:
+                _LOG.warning(
+                    "Task '%s' on quantum %s exited with partial outputs; "
+                    "considering this a qualified success and proceeding.",
```
parejkoj, line 508: That's a good way to phrase it. I do wonder if it shouldn't be `error` here, or `exception` maybe (we're allowed to log an `error` even if the task doesn't halt execution, right?), so it doesn't get lost in our massive field of warnings, but I guess that's up to DRP to decide.

python/lsst/ctrl/mpexec/singleQuantumExecutor.py
```diff
@@ -476,8 +477,39 @@ def runQuantum(
...
+                    task_node.label,
+                    quantum.dataId,
+                )
+                _LOG.warning(error, exc_info=error)
```
parejkoj, line 512: Or make this one `log.exception`? Not sure.

parejkoj (commented):
I don't know what executor AP uses, but if they use this, they'll have to reconfigure the new kwarg to be `True`, since they want the other behavior.

Check your DO NOT MERGE commit.

python/lsst/ctrl/mpexec/singleQuantumExecutor.py
TallJimbo: I don't really like ERROR or higher log messages that don't correspond to an exception or some other actual execution stoppage.  I thought about making this a hard failure, but I figured we'd be much more frustrated if it brought down processing that was otherwise working as intended.

python/lsst/ctrl/mpexec/singleQuantumExecutor.py
TallJimbo: I'm not sure if we're _allowed_ to log `error` if we don't halt, but I don't like doing it because I think there's no common mental model of what it means.

TallJimbo:
> I don't know what executor AP uses, but if they use this, they'll have to reconfigure the new kwarg to be  `True`, since they want the other behavior.

The Prompt Processing framework uses `SeparablePipelineExecutor`, and yes, @kfindeisen et all will probably want to start passing `raise_on_partial_outputs=True` at construction there after this merges.

> Check your DO NOT MERGE commit.

Yup, this needed to keep the rest of the GitHub Actions green, so our workflow is to drop it only after merging the branch for the upstream package.

kfindeisen:
> The Prompt Processing framework uses `SeparablePipelineExecutor`, and yes, @kfindeisen et all will probably want to start passing `raise_on_partial_outputs=True` at construction there after this merges.

If you're making a behavioral and/or breaking change to the API, will this be announced/documented anywhere? ATM I have no basis for a decision either way.

TallJimbo:
It's establishing behavior that I think was basically ill-defined before; RFC-958 defined two options but did not state which would be the default, IIRC, and this ticket should really be considered a belated part of that RFC's implementation.

As to whether Prompt Processing actually wants to raise when `calibrateImage` produces only partial outputs, @parejkoj has been adamant that the answer is "yes", while I'm a little less certain, but am willing to defer to the AP team on this.  (It's not that I expect the AP pipeline to be able to do much with those partial outputs, but failing a little later when a downstream task doesn't find what it needs - as DRP _often_ would in these cases - seems fine.)

I'll make a community post before I merge this in any case; there is some chance we'll want to do a bit of work on some downstream tasks as we get used to the full implications of this RFC.

python/lsst/ctrl/mpexec/singleQuantumExecutor.py
parejkoj: There are ERROR, FAILURE, EXCEPTION log types; I wish we'd used FAILURE for what you're describing, so that we can use ERROR for other things ("be louder than warning, but I (maybe) didn't stop running").

python/lsst/ctrl/mpexec/singleQuantumExecutor.py
parejkoj: In fact, we explicitly can! Per the only place I know of that we've written any such recommendation:
https://developer.lsst.io/stack/logging.html#log-levels

`ERROR: for errors that may still allow the execution to continue.`

parejkoj:
Maybe better to leave this for discussion on Community, but I do wonder whether the default for this should be to fail on partial output errors, with DRP changing it in their pipelines? I don't know what we should train the users to expect, when running pipelines themselves: "reproducible errors halt" or "reproducible errors continue and may or may not cause problems further down the line"?

TallJimbo:
> I do wonder whether the default for this should be to fail on partial output errors, with DRP changing it in their pipelines? I don't know what we should train the users to expect, when running pipelines themselves: "reproducible errors halt" or "reproducible errors continue and may or may not cause problems further down the line"?

The vast majority of reproducible errors still halt, regardless of the new option - it's just the ones that also have a potentially-useful partial output that now default to proceeding.  I think that's consistent with us wanting to process as far as we can in a lot of contexts (at least Rapid Analysis as well as DRP, and I think it's plausible alert production in special programs could go this direction, too, even if it doesn't in WFD).  This does put the burden on the downstream tasks to handle those partial outputs, and I do suspect some of them are not ready for that responsibility yet (in particular, we'll get some confusing error messages from them at first).  But I also think that's the only place that responsibility can realistically be.

python/lsst/ctrl/mpexec/singleQuantumExecutor.py
kfindeisen: Note that `exception` is not a distinct log level; it's just [`error` with automatic stack trace printing](https://docs.python.org/3/library/logging.html#logging.Logger.exception).

PR #304, opened 2024-08-23
DM-45894: Fix dimension universe instantiation in subprocesses

andy-slac:
With run-qbb when dimension universe is not the default one, we need
to make sure that that universe is instantiated from a configuration
in a subprocess before unpickling anything that requires that universe.
The fix uses butler factory to do that, using its unpickle method,
and there is some complicated logic in executors to make sure that
unpickling of that factory happens first thing in a subprocess.
There may be an easier way to enforce that but it would require
interface changes for MPGraphExecutor or SingleQuantumExecutor.

Subprocess parameters are now pickled manually to guarantee that
logging is initialized properly in a subprocess before any complex
unpickling happens.

## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/ctrl/mpexec/cmdLineFwk.py
+++ b/python/lsst/ctrl/mpexec/cmdLineFwk.py
@@ -36,6 +36,7 @@ import atexit
 import contextlib
 import copy
 import logging
+import pickle
 import shutil
 from collections.abc import Mapping, Sequence
 from types import SimpleNamespace
@@ -49,6 +50,7 @@ from lsst.daf.butler import (
     Config,
     DatasetId,
     DatasetType,
+    DimensionConfig,
     DimensionUniverse,
     LimitedButler,
     Quantum,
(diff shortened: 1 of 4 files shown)


python/lsst/ctrl/mpexec/cmdLineFwk.py
```diff
@@ -521,6 +530,27 @@ def __call__(self, quantum: Quantum) -> LimitedButler:
...
+        universe = DimensionUniverse(dimensions_config)
+        dataset_types = pickle.loads(dataset_types_pickle)
+        return _QBBFactory(butler_config, universe, dataset_types)
+
+    def __reduce__(self) -> tuple:
+        # If dimension universe is not default one, we need to dump/restore
+        # its config.
+        config = self.dimensions.dimensionConfig
+        default = DimensionConfig()
+        if (config["namespace"], config["version"]) != (default["namespace"], default["version"]):
```
timj, line 546: Maybe add a comment saying that you don't need to send the config in the pickle if the version matches the default universe?

python/lsst/ctrl/mpexec/mpGraphExecutor.py
```diff
@@ -145,14 +147,18 @@ def _executeJob(
 
         Parameters
         ----------
-        quantumExecutor : `QuantumExecutor`
-            Executor for single quantum.
-        task_node : `lsst.pipe.base.pipeline_graph.TaskNode`
-            Task definition structure.
+        quantumExecutor_pickle : `bytes`
+            Executor for single quantum, pickled.
+        task_node : `bytes`
```
timj, line 152:
```suggestion
        task_node_pickle : `bytes`
```
(I think numpydoc ignored the problem because it's a private method)

timj (approved):
Looks good.

PR #309, opened 2024-09-27
DM-46525: switch back to raising on partial output errors by default

TallJimbo:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/ctrl/mpexec/cli/opt/options.py
+++ b/python/lsst/ctrl/mpexec/cli/opt/options.py
@@ -450,9 +450,10 @@ fail_fast_option = MWOptionDecorator(
 )
 
 raise_on_partial_outputs_option = MWOptionDecorator(
-    "--raise-on-partial-outputs",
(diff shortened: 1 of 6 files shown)


python/lsst/ctrl/mpexec/cli/opt/options.py
```diff
@@ -450,9 +450,10 @@
 )
 
 raise_on_partial_outputs_option = MWOptionDecorator(
-    "--raise-on-partial-outputs",
+    "--raise-on-partial-outputs/--no-raise-on-partial-outputs",
```
parejkoj, line 453: That's cool, I didn't know you could do that.

parejkoj (approved):

python/lsst/ctrl/mpexec/cli/opt/options.py
TallJimbo: I learned about it on this ticket, too!

